PR #5, opened 2025-05-13
feat: MTA resource v1alpha1 version

Lasserich:
This PR merges the previous implementation of the MTA Provider with this CloudFoundry Provider.

**Reasoning**

This is done to adhere to the standards of the [CloudFoundry Terraform Provider](https://registry.terraform.io/providers/cloudfoundry/cloudfoundry/latest/docs/resources/mta), having chosen the same path instead of creating separate providers. Beyond this reasoning there are also synergies created by re-using the same CloudFoundry client for both APIs and enabling Provider native references of the Space resource.

**Example**
```yaml
kind: Mta
metadata:
  name: test
  namespace: default
spec:
  forProvider:
    namespace: default
    spaceRef: 
      name: my-space
    blueGreenDeploy: true
    file:
      url: <url>
      credentialsSecretRef:
        name: secret
        namespace: default
    extension: |
      _schema-version: 3.3.0
      extends: my-app
      ID: my-app.extension
      modules:
        - name: my-app-srv
          parameters:
            instances: 2
```
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret
  namespace: default
type: Opaque
stringData:
  user: my-user
  password: my-password
  cred: my-user:my-password
```

**Done criteria are:**
- [x] implement MTA client
- [x] implement MTA resource
- [x] provide working example 
- [x] unify provider config
- [ ] unify secrets (?) -> so actually we use a docker-registry type secret for our App deployments. It does not seem feasible to use that also here for .mtar types
- [x] support `space` reference parameter
- [ ] implement basic unit tests for Controller (Create, Observe, Delete) only -> Client is not managed by us so we rely on their testing etc.


SatabdiG:
Closing in favor of MTA provider: https://github.tools.sap/cloud-orchestration/crossplane-provider-mta

PR #9, opened 2025-05-15
chore: updating go-cfclient version

gergely-szabo-sap:


dwlou (approved):
lgtm

Lasserich:
I think there is something misconfigured here, since code-scanning is not even an action. I will investigate.

PR #13, opened 2025-05-23
feat: CLI POC for importing pre-existing Cloud Foundry resources

kosinamatteo:
## Description / Idea
This pull request presents an initial Proof of Concept (POC) for a Command Line Interface (CLI) designed to import pre-existing Cloud Foundry resources. The motivation behind this feature is to simplify the process of importing and managing existing resources through Infrastructure as Data (IaD), eliminating manual and error-prone importing-work. 

### ⚠️ Note
This is an early prototype and serves as a first proposal on how such a tool could be build

### DIFF
--- a/internal/cli/adapters/app.go
+++ b/internal/cli/adapters/app.go
@@ -1,0 +1,128 @@
+package adapters
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/SAP/crossplane-provider-cloudfoundry/apis/resources/v1alpha1"
+	"github.com/SAP/crossplane-provider-cloudfoundry/internal/cli/pkg/utils"
+	"github.com/SAP/crossplane-provider-cloudfoundry/internal/crossplaneimport/client"
+	res "github.com/SAP/crossplane-provider-cloudfoundry/internal/crossplaneimport/resource"
+	cfresource "github.com/cloudfoundry/go-cfclient/v3/resource"
+	v1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+)
+
+// CFApp implements the Resource interface
+type CFApp struct {
+	managedResource *v1alpha1.App
+	externalID      string
+}
+
+func (d *CFApp) GetExternalID() string {
+	return d.externalID
+}
+
+func (d *CFApp) GetResourceType() string {
+	return "app"
+}
+
+func (d *CFApp) GetManagedResource() resource.Managed {
+	return d.managedResource
+}
+
+func (d *CFApp) SetProviderConfigReference(ref *v1.Reference) {
+	d.managedResource.Spec.ProviderConfigReference = ref
+}
+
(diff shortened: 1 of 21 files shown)


internal/cli/adapters/app.go
```diff
@@ -0,0 +1,128 @@
...
+	managedResource *v1alpha1.App
+	externalID      string
+}
+
+func (d *CFApp) GetExternalID() string {
+	return d.externalID
+}
+
+func (d *CFApp) GetResourceType() string {
+	return "app"
```
dwlou, line 28: As the api package is imported, we could directly `return v1alpha1.App_Kind`.

internal/cli/adapters/client.go
```diff
@@ -0,0 +1,305 @@
...
+	if !ok {
+		return nil, fmt.Errorf("name filter is required for spaces")
+	}
+	orgName, ok := filter["org"]
+	if !ok {
+		return nil, fmt.Errorf("org-reference filter is required for spaces")
+	}
+
+	// Get all spaces from CF
+	responseCollection, err := c.client.Spaces.ListAll(ctx, &cfv3.SpaceListOptions{})
```
dwlou, line 82: If possible, instead of getting all resources and filter the results afterwards, it would be better to construct the `xxxListOptions` from the `filters` and get only the valid result back. See https://github.com/cloudfoundry/go-cfclient?tab=readme-ov-file#filtering

internal/cli/adapters/client.go
```diff
@@ -0,0 +1,305 @@
...
+
+func (c *CFClient) getOrganizations(ctx context.Context, filter map[string]string) ([]interface{}, error) {
+	// Get GUID filter
+	name, ok := filter["name"]
+	if !ok {
+		return nil, fmt.Errorf("name filter is required for organizations")
+	}
+
+	// Get organizations from CF
+	organizations, err :=  c.client.Organizations.ListAll(ctx, &cfv3.OrganizationListOptions{})
```
dwlou, line 123: pass name filter to the API: 
```go
	listOptions := cfv3.NewOrganizationListOptions()
	listOptions.Names.EqualTo(name)

	organizations, err := c.client.Organizations.ListAll(ctx, listOptions)

```

internal/cli/adapters/app.go
```diff
@@ -0,0 +1,128 @@
+package adapters
```
dwlou, line 1: how about version adapter like the apis: `/adapters/v1alpha1` so that we can deal with multiple api versions in the same provider.

internal/cli/adapters/app.go
kosinamatteo: Thank you, that looks way better. I integrated this with my recent commit.

gergely-szabo-sap:
I created a [feature branch](https://github.com/SAP/crossplane-provider-cloudfoundry/tree/feat/import-cli-tool) and a draft [PR](https://github.com/SAP/crossplane-provider-cloudfoundry/pull/63). The development can continue there.

PR #59, opened 2025-05-30
fix: Setup controller for providerconfig

### LINKED ISSUE #11: [BUG] ProviderConfig can be deleted while still having usage
**Describe the bug**
ProviderConfig can be deleted while there are still resources referencing it. This causes all the resources to be stuck (especially when applying the ProviderConfig and other resources simultaneously, e.g. with helm or kustomize)

v0.3.0

**To Reproduce**
Create any resource referencing a cloudfoundry ProviderConfig. Delete the ProviderConfig. It will be deleted.

**Expected behavior**
Deletion should be blocked until there are resources referencing it. 

**Screenshots**

**Additional context**

dwlou:
This PR fixes #11  by setting up controller for `ProviderConfig` so that it blocks deletion while there is usage. 
image


sdischer-sap (approved):
lgtm

Issue #54, opened 2025-05-28
[BUG] Cannnot reference Space by name in Service Instance


dwlou:
**Describe the bug**

When importing a serviceinstance using external without specifying space or space Ref, the reconcilation fails because of the following error

`Cannot resolve space name. The resource does not implement SpaceScoped  `

**Tested with Provider Version**
v0.3.0

**To Reproduce**

Apply the following resource

```yaml
# Import ServiceInstance with external-name (guid)
apiVersion: cloudfoundry.crossplane.io/v1alpha1
kind: ServiceInstance
metadata:
  name: my-servicenstance-external-name
  namespace: default
  annotations:
    crossplane.io/external-name: 416ed10a-220b-479f-bd32-12bc09249f16
spec:
  forProvider:
    type: managed
    name: service-instance
    servicePlan:
      offering: destination
      plan: lite
  managementPolicies:
    - Observe
```

**Expected behavior**

Should import the service instance, if GUID is valid.

dwlou:
As the error described, the ServiceInstance does not have the method `GetSpaceRef` as expected for a space scoped resource.  Service Instance is space scoped. A related problem is that currently the controller expects that service instance CR would always specify a reference to space, either by name or by x-ref and attempts to resolve the reference.   However, this assumption does not hold true for the import case as user may not specify provide the external name (guid) of the service instance.

sdomarecki-sap:
Fixed along with #60

PR #60, opened 2025-05-30
Fix: resolve space name reference only when provided

dwlou:
This PR allows user to import space-scoped resources, e.g., space role, using external-name but not directly specifying the space of the resource.  The following example is now valid. 

```yaml
# import space role with external-name (guid)
kind: SpaceRole
metadata:
  namespace: default
  name: my-spacerole-external-name
  annotations:
    crossplane.io/external-name: 55438c93-23d9-43cd-adae-bf186d6bac17
spec:
  forProvider: {}
  managementPolicies:
    - Observe
```
Previously it returns unknown space error. 
image

If neither external-name nor space is specified, then the controller will error during observe

image

### DIFF
--- a/apis/resources/v1alpha1/serviceinstance_types.go
+++ b/apis/resources/v1alpha1/serviceinstance_types.go
@@ -292,3 +292,9 @@ func (r *ServiceInstance) GetCloudFoundryName() string {
 	}
 	return *r.Spec.ForProvider.Name
 }
+
+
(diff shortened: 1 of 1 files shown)


sdischer-sap:
Relates to this issue right? 
https://github.com/SAP/crossplane-provider-cloudfoundry/issues/54

sdischer-sap (approved):
lgtm

PR #61, opened 2025-06-02
fix: ServiceInstance - reference Space by GUID

sdomarecki-sap:
This PR solves #54 
SpaceRef may be empty when external-name is provided.

The functionality is tested manually, and it successfully imports a ServiceInstance if it exists already.
Test object:
```
# Import ServiceInstance with external-name (guid)
kind: ServiceInstance
metadata:
  name: my-servicenstance-external-name
  namespace: default
  annotations:
    crossplane.io/external-name: d6d339fa-c4f7-4372-8878-4f1ef0266938
spec:
  forProvider:
    type: managed
    name: service-instance
    servicePlan:
      offering: destination
      plan: lite
  managementPolicies:
    - Observe
```
image

### DIFF
--- a/internal/clients/space/reference.go
+++ b/internal/clients/space/reference.go
@@ -26,10 +26,7 @@ func ResolveByName(ctx context.Context, clientFn clients.ClientFn, mg resource.M
 	// if external-name is not set, search by Name and Space
 	sr := cr.GetSpaceRef()
 	if sr == nil || sr.SpaceName == nil {
-		if sr.Space != nil { // space GUID is directly set, so we do not need to use names.
-			return nil
-		}
-		return errors.New("Unknown space. Please specify `spaceRef` or `spaceSelector` or using `spaceName` and `orgNames`. ")
+		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


internal/clients/space/reference.go
```diff
@@ -26,10 +26,7 @@ func ResolveByName(ctx context.Context, clientFn clients.ClientFn, mg resource.M
 	// if external-name is not set, search by Name and Space
```
dwlou, line 26: should fix comment

internal/clients/space/reference.go
```diff
@@ -26,10 +26,7 @@ func ResolveByName(ctx context.Context, clientFn clients.ClientFn, mg resource.M
...
 	sr := cr.GetSpaceRef()
 	if sr == nil || sr.SpaceName == nil {
```
dwlou, line 28: should check orgName too
 ```go 
if sr== nil || sr.SpaceName == nil || sr.OrgName == nil 
```

sdomarecki-sap:
Closing as it overlaps with #60

PR #89, opened 2025-07-01
feat: enhance ServiceCredentialBinding for Rotation

### LINKED ISSUE #87: [FEATURE] Enhance `ServiceCredentialBinding` for rotate
### Context

This issue is to implement #26 by directly enhancing `ServiceCredentialBinding` type and controller.  **The fundamental principle of this model is that there is no difference between a static key and a rotating key. A static `ServiceCredentialBinding` is simply a rotating binding with an infinite `frequency` and `ttl`.**

This view allows the controller to use a single, unified logic path for all ServiceCredentialBinding resources,  reducing complexity and eliminating the need for separate "static" and "rotating" modes.

### Scope
We found that rotating a Secret with type `app` requires to restart the app (as rolling update). It does not feel right to do that as part of the key rotation of another CR. We therefor leave it out of scope for now. 

#### 1.  Update `spec.forProvider` with `rotation` block.
```golang
type RotationParameters struct {
    // Frequency defines how often the active key should be rotated.
    Frequency *metav1.Duration `json:"frequency"`

    // TTL (Time-To-Live) defines the total time a credential is valid for before it is deleted.
    // Must be >= frequency and <= 3 * frequency.
    // +kubebuilder:validation:Optional
    // +kubebuilder:validation:CEL=expression="!has(self.ttl) || (!has(self.frequency) || (self.ttl >= self.frequency && self.ttl <= self.frequency * 3))",message="ttl must be greater than or equal to frequency and no more than 5 times frequency"
    TTL *metav1.Duration `json:"ttl,omitempty"`
}
```
The `rotation` block is entirely optional to maintain backward compatibility. 

*  **When `rotation` is defined**, the user provides explicit `frequency` and `ttl` values.
    ```yaml
    spec:
      forProvider:
         # A key that rotates every 30 days
        rotation:
          frequency: "720h"
          ttl: "1440h"
    ```
*   **When `rotation` is omitted**, the controller internally assumes an infinite `frequency` and `ttl`. The user is effectively creating a key that will never be scheduled for rotation or cleanup.

#### 2. State Management

Whether they are configured to rotate or not, the state management rules are the same.  

-   **`external-name` is always the active key**: The `metadata.annotations['crossplane.io/external-name']` annotation on the CR always points to the GUID of the single active key. For a static binding, this will be its one and only key.

-   **The Secret is always from the active key**: The user-facing secret is always populated from the credentials of the key identified in `external-name`. **The credentials of retired key are not published.**

-   `status.atProvider` is always active service credential binding plus an optional list of `retiredKeys` that keeps track of service binding resources that have been rotated out but transitionally retained due to `ttl` configuration. This changes is incremental and maintain backward compatibility.  
```golang
type ServiceCredentialBindingObservation struct {
	Resource `json:",inline"`
	// LastOperation describes the last operation performed on the service credential binding.
	LastOperation *LastOperation `json:"lastOperation,omitempty"`

	// If the binding is rotated, `retiredBindings` stores resources that have been rotated out but are still transitionally retained due to TTL setting
	// +kubebuilder:validation:Optional
	RetiredKeys []Resource `json:"retiredKeys,omitempty"` 
}
```

#### 3. The Controller's Logic

The controller's reconciliation loop should not contain an `if rotation_enabled` block, but a single control flow for rotating credential binding. 

##### **`Observe()` 

1.  We *ONLY* fetch the active key  from Cloud Foundry as annotation by `external-name` to the CR.  There is no need to fetch any retired key from Cloud Foundry. 
2. The only changes are:
      - When the active key is observed and its age the `frequency` threshold -> add the current key to the list `retiredKeys` and return observation `ResourceExists: false`,  which subsequently triggers `Create()` to create new active key.
     -  If any **retired keys** to reach `ttl`, return `ResourceUpToDate: false`, which subsequently triggers `Update()` to do clean up. Again, no need to fetch these keys from Cloud Foundry.

##### **`Create()`, `Update()`, and `Delete()`**

-   `Create()`: creates a new key, update `external-name`. 
-   `Update()`: Handles only the cleanup of expired retired keys. 
-   `Delete()`: Always deletes all associated keys.

#### 4 Drift Detection and Correction
- Scenario 1: The *Active* Cloud Foundry Key is Manually Deleted  --> Will be recreated. Rotation cycle shift according the  new timestamp
- Scenario 2: An *Inactive* Cloud Foundry Key is Manually Deleted  --> This is "benign drift." Someone has simply done the controller's cleanup job for it.
- Scenario 3: The User-Facing Secret is Manually Modified or Deleted --> Delete not possible 
- Scenario 4: The `external-name` Annotation is Tampered With -->  Will recreated. may results in Orphan resource in CF.

Noel-Ch:
This PR adds a new feature to support automatic key rotation to enhance security by regularly refreshing credentials. This feature is particularly useful for long-running applications that need to maintain secure access to services.

Rotation is done by recreating the ServiceCredentialBinding so that the secret is written with new service credentials.

```yaml
kind: ServiceCredentialBinding
metadata:
  name: my-cloud-binding
spec:
  forProvider:
    type: key
    name: my-cloud-binding
    serviceInstanceRef:
      name: my-cloud-logging
    rotation:
      frequency: 5m
      ttl: 8m
  writeConnectionSecretToRef:
    name: my-cloud-key
```

Closes #87


internal/clients/servicecredentialbinding/servicecredentialbinding.go
```diff
@@ -123,56 +127,56 @@ func GetConnectionDetails(ctx context.Context, scbClient ServiceCredentialBindin
 }
 
 // newListOptions generates ServiceCredentialBindingListOptions according to CR's ForProvider spec
-func newListOptions(spec v1alpha1.ServiceCredentialBindingParameters) (*client.ServiceCredentialBindingListOptions, error) {
+func newListOptions(cr v1alpha1.ServiceCredentialBinding) (*client.ServiceCredentialBindingListOptions, error) {
```
dwlou, line 130: Can we avoid changing the function signature?

internal/clients/servicecredentialbinding/servicecredentialbinding.go
```diff
@@ -123,56 +127,56 @@ func GetConnectionDetails(ctx context.Context, scbClient ServiceCredentialBindin
...
+		opt.Names.EqualTo(cr.Status.AtProvider.Name)
 	}
 
 	return opt, nil
 }
 
 // newCreateOption generates ServiceCredentialBindingCreate according to CR's ForProvider spec
-func newCreateOption(spec v1alpha1.ServiceCredentialBindingParameters, params json.RawMessage) (*resource.ServiceCredentialBindingCreate, error) {
-	if spec.ServiceInstance == nil {
+func newCreateOption(cr v1alpha1.ServiceCredentialBinding, params json.RawMessage) (*resource.ServiceCredentialBindingCreate, error) {
```
dwlou, line 158: Can we avoid changing the function signature?

internal/clients/servicecredentialbinding/servicecredentialbinding.go
```diff
@@ -123,56 +127,56 @@ func GetConnectionDetails(ctx context.Context, scbClient ServiceCredentialBindin
...
+	if cr.Spec.ForProvider.ServiceInstance == nil {
 		return nil, errors.New(ErrServiceInstanceMissing)
 	}
 
 	var opt *resource.ServiceCredentialBindingCreate
-	switch spec.Type {
+	switch cr.Spec.ForProvider.Type {
 	case "key":
-		if spec.Name == nil {
+		if cr.Status.AtProvider.Name == "" {
```
dwlou, line 166: `atProvider.Name` for `createOption` does not make sense. If a new key is needed (e.g., during the first reconciliation), there wouldn't be anything  at `atProvider`..

internal/controller/servicecredentialbinding/controller.go
```diff
@@ -175,17 +186,38 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.New(errWrongCRType)
 	}
 
+	if externalName := meta.GetExternalName(cr); externalName != "" {
+		keyRetired := false
```
dwlou, line 190: This check should have happened before Create is being called. In other words, if the current key is not retired yet, then `Create` should not have called in the first place.

internal/controller/servicecredentialbinding/controller.go
```diff
@@ -119,9 +123,9 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	}
 
 	guid := meta.GetExternalName(cr)
-	serviceBinding, err := scb.GetByIDOrSearch(ctx, c.scbClient, guid, cr.Spec.ForProvider)
+	serviceBinding, err := scb.GetByIDOrSearch(ctx, c.scbClient, guid, *cr)
 	if err != nil {
-		if clients.ErrorIsNotFound(err) {
+		if err.Error() == scb.ErrNameMissing || cfresource.IsResourceNotFoundError(err) || cfresource.IsServiceBindingNotFoundError(err) {
```
dwlou, line 128: We could implement CEL for such validation rules, like name is required for scb with type `key`. btw, good hint about using specific error checking `isServiceBindingNotFoundError`, which checks the error code, rather the message.  Let's how this can been used in all other resources

internal/controller/servicecredentialbinding/controller.go
```diff
@@ -175,6 +204,20 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalCreation{}, errors.New(errWrongCRType)
 	}
 
+	// Ensure that the current service key is retired before creating a new one.
+	if externalName := meta.GetExternalName(cr); externalName != "" {
```
dwlou, line 208: During initial apply, Crossplane sets external-name to metadata.name, if external-name is not provided.  We need to add a format check that the external-name is a valid uuid format

internal/controller/servicecredentialbinding/controller.go
```diff
@@ -121,13 +128,25 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	guid := meta.GetExternalName(cr)
 	serviceBinding, err := scb.GetByIDOrSearch(ctx, c.scbClient, guid, cr.Spec.ForProvider)
 	if err != nil {
-		if clients.ErrorIsNotFound(err) {
+		if errors.Is(err, cfclient.ErrExactlyOneResultNotReturned) ||
+			cfresource.IsResourceNotFoundError(err) ||
+			cfresource.IsServiceBindingNotFoundError(err) {
+			cr.SetConditions(xpv1.Unavailable().WithMessage(errGet))
+			meta.SetExternalName(cr, "") // Clear the external name if the resource does not exist
```
dwlou, line 135: We need to be aware that Crossplane sets `external-name` by default to `matadata.name` during initialisation.

apis/resources/v1alpha1/servicecredentialbinding_types.go
```diff
@@ -12,9 +12,13 @@ import (
 )
 
 type ServiceCredentialBindingObservation struct {
-	Resource `json:",inline"`
+	SCBResource `json:",inline"`
```
sdischer-sap, line 15: Why is this changed to this specific type?

internal/clients/servicecredentialbinding/servicecredentialbinding.go
```diff
@@ -78,17 +81,12 @@ func Create(ctx context.Context, scbClient ServiceCredentialBinding, forProvider
 		}
 	}
 
-	opts, err := newListOptions(forProvider)
-	if err != nil {
-		return nil, err
-	}
-	return scbClient.Single(ctx, opts)
-
+	return scbClient.Single(ctx, createToListOptions(opt))
```
sdischer-sap, line 84: Why did this change? How is it related to secret rotation?

internal/controller/servicecredentialbinding/controller.go
```diff
@@ -121,51 +128,39 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
-		if clients.ErrorIsNotFound(err) {
-			return managed.ExternalObservation{ResourceExists: false}, nil
-		}
-		return managed.ExternalObservation{}, errors.Wrap(err, errGet)
+		return c.handleGetError(ctx, cr, err)
 	}
 
 	if serviceBinding == nil {
+		cr.SetConditions(xpv1.Unavailable().WithMessage(errGet))
+		meta.SetExternalName(cr, "")
```
sdischer-sap, line 136: I am guessing this is the regular "not found" case right? 
You don't need to take care of any of that manually. Just return `ResourceExists: false`. 
This will call the Create() which will reset the conditions and eventually set external-name to a new value.

internal/controller/servicecredentialbinding/controller.go
```diff
@@ -188,6 +188,16 @@ func (c *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 
 	meta.SetExternalName(cr, serviceBinding.GUID)
 
+	if cr.ObjectMeta.Annotations != nil {
+		if _, ok := cr.ObjectMeta.Annotations[ForceRotationKey]; ok {
+			meta.RemoveAnnotations(cr, ForceRotationKey)
+		}
+	}
+
+	if err := c.kube.Update(ctx, cr); err != nil {
```
sdischer-sap, line 197: Is this manual update really required? Shouldn't crossplane call an update on its own.

sdischer-sap (changes requested):
Thanks for all the hard work. The results are impressive. 
I added a couple of remarks regarding crossplane best practices. 

Also its missing a bit of structure and most importantly unittests.

internal/clients/servicecredentialbinding/servicecredentialbinding.go
Noel-Ch: Previously, we generate the list options from `spec.forProvider` using the name of the resource, so that we can search for the resource using its name.

However, CF does not allow two service keys to have the same name, so a rotating key uses a randomly generated name. This also means that `forProvider` does not contain the necessary information to find the newly created key.

The new approach searches for the new key using its GUID, which is stored in the create options.

internal/controller/servicecredentialbinding/controller.go
```diff
@@ -236,3 +268,128 @@ func extractParameters(ctx context.Context, kube k8s.Client, spec v1alpha1.Servi
...
+	return false
+}
+
+func (c *external) handleGetError(ctx context.Context, cr *v1alpha1.ServiceCredentialBinding, err error) (managed.ExternalObservation, error) {
+	if errors.Is(err, cfclient.ErrNoResultsReturned) ||
+		errors.Is(err, cfclient.ErrExactlyOneResultNotReturned) ||
+		cfresource.IsResourceNotFoundError(err) ||
+		cfresource.IsServiceBindingNotFoundError(err) {
+		cr.SetConditions(xpv1.Unavailable().WithMessage(errGet))
+		meta.SetExternalName(cr, "") // Clear the external name if the resource does not exist
```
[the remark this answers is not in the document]
Noel-Ch: In the [current design](https://github.com/SAP/crossplane-provider-cloudfoundry/issues/87#issue-3179069574), the external-name actually serves an additional purpose: to indicate that a key is active.

For active keys, we want to prevent cases where the external-name is overwritten before the key is added to the retired list, as this can lead to orphaned keys. Flushing the external-name is currently used to indicate that it is safe to write a new key to external-name, either because it has been retired or it is missing.

But this is not clear just from reading the code, so maybe it might be worth redesigning how this works.

internal/controller/servicecredentialbinding/controller.go
Noel-Ch: I couldn't find clear documentation on when crossplane calls an update, but from looking at the [code](https://github.com/crossplane/crossplane-runtime/blob/f5f608c93ded2ef2ea30daa38a6fd7aaed53efb2/pkg/reconciler/managed/reconciler.go#L824), it seems to usually only call `client.Status().Update()` instead of `client.Update()`, except before `Create()` and when late initializing.

I'll look through the code again and do some tests, to check if there are any unnecessary updates.

sdischer-sap:
Talked to @GenosseOtt regarding required documentation. We have an issue for that now: 
https://github.com/SAP/crossplane-provider-cloudfoundry/issues/91
@Noel-Ch If you can create a small draft for that that would be great, otherwise let us know, so we can eventually deal with that.

internal/controller/servicecredentialbinding/controller.go
Noel-Ch: I looked further into this, and had a surprising finding.

> - changes in status during Observe will be saved

Not safe to assume this if Observe returns `resourceExists: false`, because `client.Update()` would be called before Create and changes to the status will be lost.

> - setting external-name in Create will be saved

Changes are saved

internal/controller/servicecredentialbinding/controller.go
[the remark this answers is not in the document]
Noel-Ch: Now that we figured out when updates to the status and external name persist/don't persist, I should be able to implement the pattern you described without the additional verifications

sdischer-sap (commented):
Very nicely done. You really like the way you use this rotator interface. Made it a pleasure to review and a lot easier to understand. 

Just added some questions. Feel free to add unittesting. I think we are really getting close to the final feature here!

internal/clients/servicecredentialbinding/servicecredentialbinding.go
Noel-Ch: When no rotation settings are defined, we handle it just like any other rotating resource, except that the rotation frequency is infinite so the resource never expires.

From the user's perspective, the only change is that when a new external resource is created, it will be created with a randomly generated name instead of name defined in the spec. This is how we handle the change:

1. Search for the resource using the GUID in the external-name. If this succeeds, manage this resource
2. If the previous step fails, search for the resource using the name defined in the spec, and manage this resource if found. This step is for backwards compatibility with old resources created before the change
3. If the previous step fails, create a new resource with a randomly generated name, and store the GUID in external-name

apis/resources/v1alpha1/servicecredentialbinding_types.go
```diff
@@ -21,6 +21,9 @@ type ServiceCredentialBindingObservation struct {
 	RetiredKeys []*SCBResource `json:"retiredKeys,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!(has(self.type) && self.type == 'app') || !has(self.rotation)",message="rotation cannot be enabled when type is app"
```
sdischer-sap, line 24: nice thinking!

internal/clients/servicecredentialbinding/servicecredentialbinding.go
```diff
@@ -242,12 +243,18 @@ const (
...
+var (
+	src      = rand.NewSource(time.Now().UnixNano())
+	srcMutex sync.Mutex
+)
 
 func randomString(n int) string {
 	sb := strings.Builder{}
 	sb.Grow(n)
 
+	srcMutex.Lock()
```
sdischer-sap, line 255: Just out of curiosity. Why is it required here to be thread safe?

sdischer-sap (commented):
Awesome work!
Very clean and to the point, everything covered for tests thats worth. 

Holding off with the approve for now until we fixed the e2e tests though.

internal/clients/servicecredentialbinding/servicecredentialbinding.go
Noel-Ch: Controller runtime supports concurrency using the [MaxConcurrentReconciles](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller#TypedOptions) option. I'm not sure if it is enabled for Crossplane, but I added the mutex just to be safe

Lasserich (commented):
LGTM incredible work 🚀 , I would like to have this functionality also be part of our e2e tests. Is it feasible with a very short rotation e.g. 1min to check if the resource stays healthy? Wdyt @Noel-Ch? Last but not least please make sure to spread the word, put the scb example yaml in the original ticket and in this PR description please.

PR #96, opened 2025-07-28
fix: make docker image updatable for apps

### LINKED ISSUE #95: [BUG] App are not reconcile for docker image update
**Describe the bug**
When the version of docker image in the spec.forProvider.docker.image is updated. The provider does not update the docker image deployed in CF accordingly. Therefore, the version of image in `spec.forProvider.docker.image` and in`status.atProvider.applicaytions.docker.image` mismatches, but the Condition is `ready`

v0.3.1

**To Reproduce**
1. Please try to deploy and app with older image of docker image
2. update the manifest with a newer version
3. observe in CF the latest update

**Expected behavior**
I expect when the version of docker image in the spec.forProvider.docker.image is updated, the conciliation starts and update the CF app immediately

**Screenshots**
N/A

**Additional context**
N/A

dattito:
fixes #95

The `spec.forProvider.docker.image` value is currently not updatable, and a change is just ignored. This PR adds the reconcilation logic. It enables GitOps workflows.

### DIFF
--- a/internal/clients/app/app.go
+++ b/internal/clients/app/app.go
@@ -139,7 +139,7 @@ func GenerateObservation(res *resource.App) v1alpha1.AppObservation {
 
 // ChangeDetection represents what fields have changed
 type ChangeDetection struct {
-	ChangedFields []string
+	ChangedFields map[string]struct{}
 }
 
 func (cd *ChangeDetection) HasChanges() bool {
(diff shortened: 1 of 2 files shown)


internal/clients/app/app.go
```diff
@@ -126,9 +140,24 @@ func GenerateObservation(res *resource.App) v1alpha1.AppObservation {
...
-	return spec.Name == status.Name
+	if spec.Lifecycle == "docker" && spec.Docker != nil {
+		// For docker apps, check if the image has changed
+		// The AppManifest in status.AtProvider contains the current image
+		appManifest, err := getAppManifest(spec.Name, status.AppManifest)
+		if err == nil && appManifest.Docker != nil {
+			if spec.Docker.Image != appManifest.Docker.Image {
+				return false
+			}
+		}
```
gergely-szabo-sap, line 151: Don't we want to handle somehow if `err != nil`?

internal/controller/app/controller.go
```diff
@@ -201,6 +200,28 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
+			if err != nil {
+				return managed.ExternalUpdate{}, errors.Wrap(err, errSecret)
+			}
+			_, err = c.client.UpdateAndPush(ctx, guid, cr.Spec.ForProvider, dockerCredentials)
+			if err != nil {
+				return managed.ExternalUpdate{}, errors.Wrap(err, errUpdateResource)
+			}
+			return managed.ExternalUpdate{}, nil
+		}
+	}
```
gergely-szabo-sap, line 222: This is strange. Why is this re-observation needed?

gergely-szabo-sap (changes requested):

internal/clients/app/app.go
dattito: yes you are right!

internal/controller/app/controller.go
dattito: you are right, it is only important to figure out what changes

internal/clients/app/app.go
```diff
@@ -123,12 +137,59 @@ func GenerateObservation(res *resource.App) v1alpha1.AppObservation {
...
 
-// IsUpToDate checks whether current state is up-to-date compared to the given
-// set of parameters.
-func IsUpToDate(spec v1alpha1.AppParameters, status v1alpha1.AppObservation) bool {
-	// rename or update ssh setting
-	return spec.Name == status.Name
+// ChangeDetection represents what fields have changed
+type ChangeDetection struct {
+	ChangedFields []string
+}
```
gergely-szabo-sap, line 143: No need to change this, but this is not the most efficient implementation of a set type in Go.

map[string]struct{} type is a more efficient and simpler alternative.

gergely-szabo-sap (changes requested):

internal/clients/app/app.go
dattito: you are right, this is an quick and easy optimization, I'll implement it.

gergely-szabo-sap (approved):
Nicely done, @dattito!

Issue #90, opened 2025-07-03
[FEATURE] Support active update of service instance parameters in all cases


dwlou:
### Background
The controller needs to update service parameters if they differ between the desired state (as defined in the Custom Resource) and the actual state in Cloud Foundry. 

There are two scenarios where the service parameters may differ:
1. **Active Update:** The user updates the desired state in the Custom Resource (CR) to change service parameters. We call this active update. 
2. **Out-of-Band Drift:** The actual parameters in Cloud Foundry drifted due to out-of-band (not via the CR) changes. 

However, not all service instances support retrieving the current parameters (actual state) from Cloud Foundry. This is an additional factor we need to consider, as the technical solution to reconcile the service parameter differ, depending on whether it is supported not. 

|                                 | **Can Retrieve Actual State**                                   | **Cannot Retrieve Actual State**                                 |
|---------------------------------|:---------------------------------------------------------------:|:----------------------------------------------------------------:|
| **Active Update**       | **1A.** Compare desired vs. actual; update if different.<br>(Direct diff possible) | **1B.** Compare desired vs. last applied; update if different.<br>(Use checksum/annotation) |
| **Out-of-Band Drift**   | **2A.** Compare desired vs. actual; update if different.<br>(Direct diff possible) | **2B.** Cannot detect drift <br>(cannot support; out of scope) |

Currently the controller supports on 1A and 2A. This issue is about adding support for 1B so that we handle active update in both cases where parameter retrieval is supported and not supported. 

### Technical Approach
- Compute a checksum (e.g., SHA256 hash) of the desired parameters from the CR.
- Store the last applied checksum in the CR’s status or as an annotation.
- On each reconciliation:
  - If parameter retrieval is supported, fetch actual parameters from Cloud Foundry and compare to desired.  (checksum is not used)
  - If not supported, compare the checksum of the desired parameters to the stored value.
- If a difference is detected, update the service parameters in Cloud Foundry and update the stored checksum.

### Acceptance Criteria
- [ ] When the user updates the service parameters in the CR, the controller detects the change. 
- [ ] The controller updates the parameters in Cloud Foundry to match the new desired state.
- [ ] The solution handles both cases where parameter retrieval is supported and not supported:
  - If supported, compare desired and actual parameters.
  - If not supported, compare desired parameters to the last applied parameters (tracked via checksum or annotation).
 
### Additional Notes
- Consider how to handle sensitive parameters and avoid logging secrets.
[...]

Issue #48, opened 2025-05-23
[BUG] credentials should be hidden in `status.atProvider`


GenosseOtt:
**Describe the bug**

Currently credentials of serviceinstance is shown in the CR status. This should be hidden.

**Tested with Provider Version**
v2.2.5 inner-source

**Screenshots**

<img width="694" alt="Image" src="https://github.com/user-attachments/assets/ba6eb068-3133-416d-b294-a1f6a291ec2c" />

**Technical Steps**
- we can annotate this field to be invisible that would be a quick fix
- we need to cache what was previously applied

**Expected behavior**

Credentials should be hidden. 

## Some background
There is no reliable way to retrieve service parameters or credentials, we need a mechanism to remember what the  last applied parameter and credential, so that the controller is able to **detect if there are changes to parameter and credentials**.   `status.atProvider.credentials` for this purpose only . 

## Solution
- if we can annotate this field to be `invisible` that would be a quick fix. Or we need a new implementation for the above problem
- we are not sure if this exists - would we have to build this? **Create follow-up if needed**

GenosseOtt:
Please provider Steps to reproduce and Expected behavior!

GenosseOtt:
Some background: There is no reliable way to retrieve service parameters or credentials, we need a mechanism to remember what the  last applied parameter and credential, so that the controller is able to detect if there are changes to parameter and credentials.   `status.atProvider.credentials` for this purpose only . 

Solution: if we can annotate this field to be `invisible` that would be a quick fix. Or we need a new implementation for the above problem

Lasserich:
@dattito maybe you can take a look here? Lot of missing information so should be mostly some research. Just un-assign if not feasible in your next few days. :)

Issue #47, opened 2025-05-23
[BUG] Change secret referenced by ServiceInstance (parameter or credentials) should trigger update of ServiceInstance


GenosseOtt:
**Describe the bug**
ServiceInstance can reference a k8s secret (`paramsSecretRef` and `credentialSecretRef`) for service configuration. If the secret is updated, the service instance should be updated with the parameters/secrets 

**Tested with Provider Version**
v2.2.5 

**To Reproduce**
1. Crate a service instance with reference to a secret: 
```yaml
apiVersion: cloudfoundry.btp.orchestrate.cloud.sap/v1alpha1
kind: ServiceInstance
metadata:
  namespace: default
  name: my-ups
spec:
  forProvider:
    type: user-provided
    name: my-ups
    routeServiceUrl: https://my-demo-route-service.example.com
    syslogDrainUrl: syslog-tls://example.log-aggregator.com:6514
    spaceRef: 
      name: my-space
      policy:
        resolve: Always
    credentialsSecretRef: 
        name: dynatrace-credentials
        namespace: default

---  
apiVersion: v1
kind: Secret
metadata:
  name: dynatrace-credentials
  namespace: default
type: Opaque
stringData:
  environmentid: environmentid
  apitoken: apitoken
  apiurl: apiurl
  type: dynatrace
```
2. Wait until `my-ups` is synced and ready. The service credential reflects the content of the secret.
3.  Chang sth in the secret
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: dynatrace-credentials
  namespace: default
type: Opaque
stringData:
  environmentid: environmentid-NEW #changed 
  apitoken: apitoken
  apiurl: apiurl
  type: dynatrace
```
4. `my-ups` is shown as synced and ready, but when check that the service credential, it is not updated and still the OLD content. 

**Expected behavior**
When reference secret changes, the controller should update/reconfigure the service instance  during the next reconciliation so that the service credential is synced with the new content of the secret.

Lasserich:
We leave this one open until @dwlou has confirmed this is not an issue anymore.

PR #98, opened 2025-07-30
fix: remove potential confidential data from ServiceInstance spec

dattito:
This PR is related to #48, #47 and #90.

With this change, the  (possible) confidential data stored in `status.atProvider.credentials` is now hidden, by storing only the hashed value, which can then be used for drift detection.

Maybe it is a now also a good idea to rename the `status.atProvider.credentials` field to avoid confusion, but IDK if this could be a breaking change, so I left it as is.

The feature request of #90 (implementing the 1B case mentioned in it's description) was already implemented(?) at least the code indicates this...

### DIFF
--- a/internal/controller/serviceinstance/controller.go
+++ b/internal/controller/serviceinstance/controller.go
@@ -175,7 +175,7 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	case v1alpha1.LastOperationSucceeded:
 		// If the last operation succeeded, set the CR to available
 		cr.SetConditions(xpv1.Available())
-		credentialsUpToDate := true
+		var credentialsUpToDate bool
 		desiredCredentials, err := extractCredentialSpec(ctx, c.kube, cr.Spec.ForProvider)
 		if err != nil {
 			return managed.ExternalObservation{}, errors.Wrap(err, errSecret)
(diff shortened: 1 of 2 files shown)


internal/controller/serviceinstance/controller.go
```diff
@@ -175,21 +175,29 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 				return managed.ExternalObservation{ResourceExists: true}, errors.Wrap(err, errGetParameters)
 			}
-			cr.Status.AtProvider.Credentials = cred
+			cr.Status.AtProvider.Credentials = iSha256(cred)
+
+			credentialsUpToDate = jsonContain(cred, desiredCredentials)
+		} else {
+			desiredHash := iSha256(desiredCredentials)
+
+			credentialsUpToDate = bytes.Equal(desiredHash, cr.Status.AtProvider.Credentials)
```
gergely-szabo-sap, line 196: Maybe I misunderstood this.

I assume that when `EnableParameterDriftDetection` is false, credential drift shall not imply an update of the external resource. But if we set `credentialsUpToDate` here, we may end up in an Update operation, see line 200.

gergely-szabo-sap (changes requested):

internal/controller/serviceinstance/controller.go
dattito: Maybe @dwlou can help clarifying this

dwlou: `EnableParameterDriftDetection` should not prevent active update, that is, if user changes the parameters in the CR, then it should perform an update.

gergely-szabo-sap (approved):
lgtm

PR #103, opened 2025-08-28
fix: fixing crash for external-name imported ServiceInstances

gergely-szabo-sap:


gergely-szabo-sap:
Service plan is required for managed service instance, unless it is using external-name for import. If this is missing, the controller shall return an error. @dwlou.

dwlou:
> Service plan is required for managed service instance, unless it is using external-name for import. If this is missing, the controller shall return an error. @dwlou.

We need to also make sure that we support update of service-plan for existing resources.  This means we need to resolve service plan GUID at every reconciliation.

internal/controller/serviceinstance/controller.go
```diff
@@ -387,22 +389,30 @@ func (s servicePlanInitializer) Initialize(ctx context.Context, mg resource.Mana
...
 
-	//  Lookup service plan during every reconciliation to detect an update service_plan of existing service.
-	cf, err := clients.ClientFnBuilder(ctx, s.kube)(mg)
-	if err != nil {
-		return errors.Wrapf(err, errNewClient)
-	}
-	opt := client.NewServicePlanListOptions()
-	opt.ServiceOfferingNames.EqualTo(*cr.Spec.ForProvider.ServicePlan.Offering)
-	opt.Names.EqualTo(*cr.Spec.ForProvider.ServicePlan.Plan)
+	if _, ok := mg.GetAnnotations()["crossplane.io/external-name"]; !ok {
```
dwlou, line 392: The service plan lookup must run all the time regardless whether "external-name" is set or not.  We need this logic to support updating service plan for existing service instances. This is not support now as the look up is omitted if external-name is provided.  To support both case, we should call `cf.ServicePlans.Single()`  whenever valid input is available.

gergely-szabo-sap: After discussing the topic with @dwlou, I got a better understanding of the goal of this method. See the fix in https://github.com/SAP/crossplane-provider-cloudfoundry/pull/103/commits/526632f1254e49431d67d1c4054768d56c935a04

dwlou (approved):
lgtm

PR #129, opened 2025-09-16
fix[114]: fixing ServiceInstance creation without credentials

### LINKED ISSUE #114: [BUG] `credentials` should be optional for `ServiceInstance` with type `user-provided`
**Describe the bug**
Currently, `credential` is a required field for user-provided service. If not provided, it  returns error `CannotCreateExternalResource  5s (x5 over 7s)  managed/serviceinstance.cloudfoundry.crossplane.io  cannot create ServiceInstance in Cloud Foundry: Missing or invalid credentials``

According to [API spec](https://v3-apidocs.cloudfoundry.org/version/3.185.0/#create-a-service-instance), so it must be allowed to create UPS without supplying a credential.

**Current behavior**
`credential` is required.

**Expected behavior**
User should allowed to create UPS without providing `credential` - optional.

mpechkurov:
This PR fixes #114 

ServiceInstance with type user-provided should be possible create without credentials.

### DIFF
--- a/internal/clients/serviceinstance/serviceinstance.go
+++ b/internal/clients/serviceinstance/serviceinstance.go
@@ -152,7 +152,7 @@ func (c *Client) Create(ctx context.Context, spec v1alpha1.ServiceInstanceParame
 		return c.createManaged(ctx, spec, creds)
 
 	case v1alpha1.UserProvidedService:
-		return c.createUserProvided(ctx, spec, creds)
+		return c.createUserProvided(ctx, spec)
 	default:
 		return nil, errors.New("unknown service instance type")
 	}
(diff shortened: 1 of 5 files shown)


internal/clients/serviceinstance/serviceinstance.go
```diff
@@ -204,7 +199,6 @@ func (c *Client) createUserProvided(ctx context.Context, spec v1alpha1.ServiceIn
 
 	// workaround: cf-goclient supports few ups options at creation time.
 	upt := resource.NewServiceInstanceUserProvidedUpdate().
-		WithCredentials(creds).
```
Lasserich, line 207: Wouldn't remove this fully. Now we take away the functionality to supply credentials all together, no? 
I think we only want to make it optional so that when the supplied cred is empty we do not throw an error.

Lasserich (changes requested):

Issue #112, opened 2025-09-05
[BUG] ServiceInstance does not delete itself when referenced secret is deleted


Rand0mF:
**Describe the bug**
Given a ServiceInstance with a secret ref `spec.forProvider.credentialsSecretRef`, deleting the secret and then the ServiceInstance (or at the same time), the ServiceInstance is stuck with the error message

```
Warning  CannotObserveExternalResource  11s (x6 over 41s)  managed/serviceinstance.cloudfoundry.crossplane.io  cannot resolve secret reference: Secret "test-ups-secret" not found
```

However, for deletion the secret would not be required. This is especially a problem if Secret (or ExternalSecret) and ServiceInstance are part of the same kustomization/helm chart.

**Tested with Provider Version**
v0.3.1

**To Reproduce**
kubectl apply and delete something like this:

```
apiVersion: v1
kind: Secret
metadata:
  name: test-ups-secret
  namespace: default
stringData:
  credentials.json: |
    {
      "some": "thing"
    }

---
apiVersion: cloudfoundry.crossplane.io/v1alpha1
kind: ServiceInstance
metadata:
  name: test-ups
spec:
  forProvider:
    name: test-ups
    type: user-provided
    credentialsSecretRef:
      name: test-ups-secret
      key: credentials.json
      namespace: default
    spaceRef:
      name: my-space
```

**Expected behavior**
ServiceInstance should be able to delete itself even if the credentials ref is gone

**Screenshots**

**Additional context**

PR #139, opened 2025-09-25
fix[112]: service instance deletion without secret

mpechkurov:
Fix bug #112

### DIFF
--- a/test/e2e/cloudfoundry_services_test.go
+++ b/test/e2e/cloudfoundry_services_test.go
@@ -56,7 +56,7 @@ func TestCloudFoundryServices(t *testing.T) {
 		},
 	)
 
-	var steps = [...]string{"space", "service_instance", "ups", "scb_key"}
+	var steps = [...]string{"space", "service_instance", "ups", "ups_no_credentials", "scb_key"}
 	for _, name := range steps {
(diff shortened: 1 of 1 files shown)


internal/controller/serviceinstance/controller.go
```diff
@@ -179,7 +179,7 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		var credentialsUpToDate bool
 		desiredCredentials, err := extractCredentialSpec(ctx, c.kube, cr.Spec.ForProvider)
 		if err != nil {
-			return managed.ExternalObservation{}, errors.Wrap(err, errSecret)
+			return managed.ExternalObservation{}, checkDeletion(cr, errors.Wrap(err, errSecret))
```
dwlou, line 182: to check deletion, we can use crossplane [`meta.WasDeleted(managed)`](https://github.com/crossplane/crossplane-runtime/blob/main/pkg/meta/meta.go#L261)

dwlou:
> See if resource exist remote in any state and no matter the LastOperation return resourceExist: true. No need to determine LastOperationState, or secret resolving etc.

That's right. If there is a ongoing deletion,  we can return an observation without checking the lastoperation.

PR #142, opened 2025-09-30
chore: add renovate

enrico-kaack-comp:
Use renovate for dependency updates. Copied from crossplane-provider-btp.
We could also use dependabot but renovate is a bit more flexible when it comes to more advanced version matching (we use in provider-btp, maybe we also need this here).


enrico-kaack-comp:
Lets use dependabot for now #143  until we have the need for more advanced configuration

PR #143, opened 2025-09-30
chore: enable dependabot for version updates

enrico-kaack-comp:
Add dependabot for package updates.
Since we dont need any specific renovate features, we can stick with dependabot


Lasserich (approved):
lgtm

Issue #214, opened 2026-03-03
[BUG] Only setting forProvider.servicePlan.id on a managed ServiceInstance causes a nil pointer exception


AndresNico:
**Describe the bug**
Creating a service instance with type = managed requires defining a servicePlan. Setting only servicePlan.id instead of servicePlan.offering and servicePlan.plan causes a nil pointer exception

**Tested with Provider Version**
v0.3.3

**To Reproduce**
Examplary resource that will cause the issue

```
apiVersion: cloudfoundry.crossplane.io/v1alpha1
kind: ServiceInstance
metadata:
  name: my-service-instance
spec:
  forProvider:
    type: managed
    name: my-service-instance
    spaceRef:
      name: my-space
      policy:
        resolve: Always
    servicePlan:
      id: f3e795d4-24cf-4823-b13b-60d3fbcec2ad
```

**Expected behavior**
That should create a ServiceInstance resource with the servicePlan set to id = f3e795d4-24cf-4823-b13b-60d3fbcec2ad (that is the UUID representing offering = destination, plan = lite)

PR #215, opened 2026-03-03
fix[214]: Update ServiceInstance controller to accept servicePlan.id only

AndresNico:
Closes: #214 
Update ServiceInstance controller to also accept servicePlan.id only


internal/controller/serviceinstance/controller.go
```diff
@@ -380,6 +380,8 @@ type servicePlanInitializer struct {
 }
```
SatabdiG: Could we also add a test for bug?

Issue #168, opened 2025-11-06
[FEATURE] 5 Add Batch 2 Resources to Upgrade Tests


SatabdiG:
## Add Batch 2 Resources to Upgrade Tests

**Part of Epic:** #163  
**Depends on:** #166  
**Can run parallel to:** #167  
**Related:** #165 (Resource categorization)

### Context

Batch 1 validates simple Space-scoped resources. Batch 2 adds medium-complexity resources including service workflows and member management. These resources have more dependencies and may require additional setup (service brokers, test users).

### Goals

- Add Batch 2 resources identified in #165
- Test service-related workflows (ServiceInstance → ServiceCredentialBinding)
- Test member management (SpaceMembers)
- Validate resources with external dependencies work through upgrade cycle
- Monitor test duration (may increase to 15-25 minutes)

### Batch 2 Resources (from #165)

**Service Workflow:**
- **ServiceInstance** - Requires Space + available service plan
- **ServiceCredentialBinding** - Requires ServiceInstance

**Member Management:**
- **SpaceMembers** - Requires Space + valid User GUID

**Total:** 3 resources with sequential dependencies

### Tasks

**Prerequisites:**
- [ ] Verify CF environment has available service brokers
- [ ] Identify service plan to use for testing (e.g., `xsuaa` lite plan)
- [ ] Obtain or document how to get test User GUID for SpaceMembers
- [ ] Update documentation with any new prerequisites

**CR Creation:**
- [ ] Create `test/upgrade/crs/serviceinstance/serviceinstance.yaml`
- [ ] Create `test/upgrade/crs/servicecredentialbinding/binding.yaml`
- [ ] Create `test/upgrade/crs/spacemembers/spacemembers.yaml`
- [ ] Ensure proper dependency ordering (ServiceInstance before Binding)

**Testing:**
- [ ] Run `make test-upgrade` locally with each new resource
- [ ] Verify ServiceInstance provisions successfully
- [ ] Verify ServiceCredentialBinding creates credentials
- [ ] Verify SpaceMembers adds user to space
- [ ] Verify all resources survive upgrade
- [ ] Verify cleanup works (especially service deletion)
- [ ] Monitor test duration 

**Documentation:**
- [ ] Update `test/upgrade/README.md` with Batch 2 resources
- [ ] Document service broker requirements
- [ ] Document User GUID requirement for SpaceMembers
- [ ] Add troubleshooting for service-related issues

### Acceptance Criteria

[...]

PR #217, opened 2026-03-06
feat[168]: Add batch 2 upgrade test resources

AndresNico:
Closes: #168 

## Summary
- Added ServiceInstance, ServiceCredentialBinding, SpaceMembers resources
- Added space to import
- Changed SpaceRole username to something less ambiguous
- Added resources + space import prerequisite to docs
- Added custom assessment logic to base test
- Local test duration < 3 min
- Resource cleanup worked

## Space Import
Space-bound resources like ServiceInstance require the technical user executing the request having the developer role on the target space. Instead of assigning that role on the created space during the test runtime I created a space, assigned the technical user the role and included it in the import


SatabdiG (approved):
🚀

PR #219, opened 2026-03-09
feat: Add service instance sharing

### LINKED ISSUE #209: Support for CF instance sharing
**Context**
We heavily use CF instance shareing to reduce cost for developer deployments

**Scope**
I want to be able to create a CF `ServiceInstance` that is actually a shared instace (probably via some `instanceRef` field

**Workarounds & Alternatives**
manually share instances 

**Additional context**
https://docs.cloudfoundry.org/devguide/services/sharing-instances.html

lennardrother:
## Summary

Closes #209

Adds support for [CF service instance sharing](https://docs.cloudfoundry.org/devguide/services/sharing-instances.html), allowing a `ServiceInstance` to be shared with additional Cloud Foundry spaces via the new `sharedSpaces` field.

- **API**: Added `sharedSpaces []SpaceReference` to `ServiceInstanceParameters`, allowing users to declare which spaces a service instance should be shared with
- **Client**: Added `AreSharedSpacesUpToDate`, `UpdateSharedSpaces`, and helper functions to compute and reconcile the diff between desired and current shared spaces
- **Controller**: Integrated shared space reconciliation into the observe, create, and update flows
- **Tests**: Unit tests for client-level sharing logic (diff, share, unshare) and controller-level reconciliation; E2E test coverage for shared service instances
- **Example**: Added `examples/serviceinstance/service_instance_shared.yaml`


internal/clients/serviceinstance/serviceinstance.go
```diff
@@ -338,3 +341,92 @@ func IsUpToDate(in *v1alpha1.ServiceInstanceParameters, observed *resource.Servi
...
+			return errors.Wrap(err, "cannot unshare service instance from spaces")
+		}
+	}
+
+	return nil
+}
+
+// getCurrentSharedSpaces retrieves the GUIDs of spaces a service instance is shared with
+func (c *Client) getCurrentSharedSpaces(ctx context.Context, guid string) ([]string, error) {
+	relationships, err := c.GetSharedSpaceRelationships(ctx, guid)
```
SatabdiG: Maybe we should handle nil return here, if c.GetSharedSpaceRelationships(ctx, guid) returns nil, never been shared, the following range will panic

SatabdiG (commented):
Great work 🚀 , mostly nits and some comments

PR #225, opened 2026-03-16
fix: separate org and space role caches

gergely-szabo-sap:
Split single roleCache/userCache into orgRoleCache/orgUserCache and spaceRoleCache/spaceUserCache to prevent data contamination between GetOrgRoles and GetSpaceRoles functions.

Also fixed cache check in GetOrgRoles to use AND logic instead of OR.


SatabdiG (approved):
🚀

PR #252, opened 2026-04-08
SpaceRole External-Name Handling

### LINKED ISSUE #251: [TASK] `External Name` (SpaceRole): Ensure ADR Compliance
**Part of Epic: #236** 

## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

AndresNico:
Closes: #251 

3. Created E2E Import Test
    - Added workflow step to set BUILD_ID env var and include it as variable
    - Added utils for E2E Import Test
5. Created Upgrade Test for External-Name Behavior
6. Added documentation


internal/controller/spacerole/controller.go
```diff
@@ -212,8 +231,12 @@ func (c *external) Delete(ctx context.Context, mg resource.Managed) (managed.Ext
 	}
 
 	// Delete is async and we need to implement wait for deletion
-	jobGUID, err := c.role.Delete(ctx, *cr.Status.AtProvider.ID)
+	jobGUID, err := c.role.Delete(ctx, meta.GetExternalName(cr))
```
SatabdiG, line 234: Since Delete now correctly uses `meta.GetExternalName(cr)`, The code above this 

```go
if cr.Status.AtProvider.ID == nil {
    return managed.ExternalDelete{}, nil
}
```
now guards on the wrong signal. It doesn't check for external name annotation which is our source of truth. If external name is set but ID is nil, not yet synced, it'll skip deletion. We should guard on `meta.GetExternalName(cr) == ""`

internal/controller/spacerole/controller.go
```diff
@@ -125,9 +126,33 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
+		resourceLateInitialized = true
+	}
+
 	guid := meta.GetExternalName(cr)
-	r, err := role.GetSpaceRole(ctx, c.role, guid, cr.Spec.ForProvider)
+
+	// ADR Step 2: External-name is set, check its format (must be valid GUID)
+	if !clients.IsValidGUID(guid) {
+		return managed.ExternalObservation{}, errors.New(fmt.Sprintf("external-name '%s' is not a valid GUID format", guid))
+	}
```
SatabdiG, line 153: I am worried this will cause a hard failure and cause the reconciler to keep retrying forever with a backoff. Should we consider setting a `Synced: False` so its more visible to the user?

internal/controller/spacerole/controller_test.go
```diff
@@ -467,3 +507,109 @@ func TestCreate(t *testing.T) {
...
+		kube    k8s.Client
+	}{
+		"SuccessfulDelete": {
+			args: args{
+				mg: fakeSpaceRole(withExternalName(guidRole), withID("resource-id")),
+			},
+			want: want{
+				mg:  fakeSpaceRole(withExternalName(guidRole), withID("resource-id")),
+				obs: managed.ExternalDelete{},
+				err: nil,
```
SatabdiG, line 541: `TestDelete` `SuccessfulDelete` is missing `withConditions(xpv1.Deleting())` in `want.mg`, but `TestDelete` doesn't `check want.mg`  — so wouldn't the Deleting condition set by the controller never be verified here?

test/e2e/import_utils.go
```diff
@@ -0,0 +1,135 @@
...
+	return ct, err
+}
+
+// DeleteResourcesIgnoreMissing deletes resources defined in a certain directory relative to testdata/crs/
+func DeleteResourcesIgnoreMissing(ctx context.Context, t *testing.T, cfg *envconf.Config, manifestDir string, timeout wait.Option) context.Context {
+	klog.V(4).Info("Attempt to delete previously imported resources")
+	r, _ := GetResourcesWithRESTConfig(cfg)
+	objects, err := test.GetObjectsToImport(ctx, cfg, []string{manifestDir})
+	if err != nil {
+		t.Fatal(objects)
```
SatabdiG: Shouldn;t this log the err instead of the slice?

test/e2e/import_utils.go
```diff
@@ -0,0 +1,135 @@
...
+			condition,
+			condition.Status == v1.ConditionTrue,
+		)
+		return result
+	})
+
+	err := wait.For(match, opts...)
+
+	if err != nil {
+		t.Error(err)
```
SatabdiG: Shoud this be `t.Fatal` instead of `t.Error`, if it were Error it would need to stop immediately but the test here continues as normal

SatabdiG (commented):
Nicely done, some reviews from my side

internal/controller/spacerole/controller.go
AndresNico: I think synced is automatically set by the Crossplane reconciler based on the result of the reconciliation loop. As we throw an error here if the GUID is not valid the resource should already get `Synced: False` but we can also set it explicitly if you prefer.

internal/controller/spacerole/controller.go
AndresNico: True. Also changed that on Observes last return statement

internal/controller/spacerole/controller.go
SatabdiG: No you are right, lets only put it in if there is specific need.

SatabdiG (approved):
🚀

PR #254, opened 2026-04-13
OrgRole External-Name Handling

### LINKED ISSUE #243: [TASK] `External Name` (OrgRole): Ensure ADR Compliance
**Part of Epic: #236** 

## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

AndresNico:
Closes: #243 

3. Created E2E Import Test
6. Created Upgrade Test for External-Name Behavior
7. Added documentation

### DIFF
--- a/internal/controller/orgrole/controller.go
+++ b/internal/controller/orgrole/controller.go
@@ -169,7 +169,7 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	cr.Status.SetConditions(xpv1.Available())
 
 	return managed.ExternalObservation{
-		ResourceExists:          meta.GetExternalName(cr) != "",
+		ResourceExists:          true,
 		ResourceUpToDate:        true,
 		ResourceLateInitialized: resourceLateInitialized,
 	}, nil
--- a/internal/controller/orgrole/controller.go
+++ b/internal/controller/orgrole/controller.go
@@ -2,6 +2,7 @@ package orgrole
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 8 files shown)


internal/controller/orgrole/controller.go
```diff
@@ -139,17 +165,11 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
-		resourceLateInitialized = true
-	}
-
 	cr.Status.AtProvider = role.GenerateOrgRoleObservation(r)
 	cr.Status.SetConditions(xpv1.Available())
 
 	return managed.ExternalObservation{
-		ResourceExists:          cr.Status.AtProvider.ID != nil,
+		ResourceExists:          meta.GetExternalName(cr) != "",
 		ResourceUpToDate:        true,
```
SatabdiG, line 173: At is point isn't meta.GetExternalName(cr) guaranteed to be true?

internal/controller/orgrole/controller.go
```diff
@@ -125,9 +126,34 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		return managed.ExternalObservation{}, errors.New(errWrongKind)
 	}
 
-	// Fetch the role object using the CloudFoundry API by guid or according to the specified parameters
+	resourceLateInitialized := false
+
+	// ADR Step 1: Check if external-name is empty
+	if meta.GetExternalName(cr) == "" {
```
SatabdiG, line 132: Crossplane sets the external-name annotation to cr.GetName() by default when no annotation is present. So a freshly created OrgRole without an explicit external-name annotation will have meta.GetExternalName(cr) == cr.GetName(), should we also check for this here or it's will fall through to valid GUID and get stuck there?

test/e2e/cloudfoundry_orgrole_import_test.go
```diff
@@ -0,0 +1,42 @@
+//go:build e2e
+
```
SatabdiG, line 2: Should we add a post-import assertion that external-name is a valid GUID?

SatabdiG (commented):
Nice work, some comments from my side

internal/controller/orgrole/controller.go
AndresNico: I think that only happens with the default initializer but we are not using that one so external-name should not be set automatically.

https://github.com/SAP/crossplane-provider-cloudfoundry/blob/900332f896967c6305cf73ab783078001d4a396d/internal/controller/orgrole/controller.go#L60-L63

If it would still get set it would fail on the GUID verification unless it is actually a valid one. In that case we end up with either
 - a resource with metadata.name == external-name annotation (if it already exists)
 - metadata.name == some GUID and external-name annotation == actual GUID from CF

So overall I think that should be fine

test/e2e/cloudfoundry_orgrole_import_test.go
AndresNico: Currently it would fail while importing when waitForResource times out as the managed resource would never get ready. But as we created the resource in the previous step, store the external-name and try to import it with that I think its very unlikely that the external-name is not a valid GUID as it is based on the response from CF. 

If you want we can add a check for that but then we got to also have potential compound keys used with other resources in mind.

SatabdiG: No, I think that's fine, you are right, CF generating a non-valid GUID would be superfluous.

internal/controller/orgrole/controller.go
SatabdiG: But wouldn't the case say the a user creates an OrgRole with a normal k8s name like "my-org-role" and no explicit external-name annotation. The controller would error out on every reconcile and never reach Create. It would be stuck. The user won't always name their orgrole resources with GUIDs.

AndresNico: If the user does not set the external-name annotation it will remain unset as the initializer is not doing that. 
https://github.com/SAP/crossplane-provider-cloudfoundry/blob/900332f896967c6305cf73ab783078001d4a396d/internal/controller/orgrole/controller.go#L252-L269

So `if meta.GetExternalName(cr) == ""` will evaluate to true and we either find the orgrole based on the spec and set the external-name, we do not find it and return `ResourceExists: false` or we encounter an error on the FindOrgRole call and return that

SatabdiG: You are right, my bad, this would only apply to controllers with Crossplane's built-in defaulting initializer.

SatabdiG (approved):
:rocket

PR #267, opened 2026-04-28
Domain External-Name Handling

### LINKED ISSUE #239: [TASK] `External Name` (Domain): Ensure ADR Compliance
**Part of Epic: #236** 

## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

Kukkerem:
Closes: #239 

3. Created E2E Import Test
4. Created Upgrade Test for External-Name Behavior
5. Added documentation


internal/clients/domain/domain.go
```diff
@@ -24,11 +25,31 @@ type Client interface {
 // Resource is the type that implements the resource.Resource interface for a Domain.
 type Resource resource.Domain
 
+// ClientWrapper wraps the domain Client.
+type ClientWrapper struct {
+	Client
+}
+
+// FindDomainBySpec looks up a domain by name when external-name is empty.
+// Name-only lookup: shared domains have no org, so org filter would exclude them.
```
SatabdiG, line 34: I really like this comment, it gives good justification!

internal/clients/domain/domain_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	wrapper := &ClientWrapper{Client: mockClient}
+	result, err := wrapper.FindDomainBySpec(ctx, spec)
+	require.NoError(t, err)
+	assert.Equal(t, expectedDomain, result)
+	mockClient.AssertExpectations(t)
+}
+
+func TestClientWrapper_FindDomainBySpec_WithOrg(t *testing.T) {
+	ctx := context.Background()
```
SatabdiG, line 85: nitpick: Do you think the name is a little misleading? `FindDomainBySpec` ignores org, so maybe `FindDomainBySpec_OrgIsIgnored` would be a better pick?

internal/controller/domain/controller_test.go
```diff
@@ -85,143 +134,151 @@ func TestObserve(t *testing.T) {
...
-					&fake.NewDomain().SetName(name).SetGUID(guid).Domain,
-					nil,
-				)
-				return m
-			},
-			kube: &test.MockClient{},
+			service: func() *mockDomainService {
+				return &mockDomainService{
+					GetDomainByGUIDFunc: func(ctx context.Context, guid string) (*cfresource.Domain, error) {
+						return nil, errors.New("CF-ResourceNotFound: The domain was not found")
```
SatabdiG: Should we use clients.ErrorIsNotFound(err) here instead of the typed string?

SatabdiG (dismissed):
Nice work

internal/clients/domain/domain.go
Kukkerem: Thank you!

internal/clients/domain/domain_test.go
Kukkerem: Absolutely right, changed!

internal/controller/domain/controller_test.go
Kukkerem: True, changed as well, thank you for the review!

SatabdiG (approved):

PR #268, opened 2026-04-28
Fix e2e Import Test: Change to dedicated users for orgrole and spacerole import test

AndresNico:
Currently both the orgrole and spacerole resource in the e2e import test refers to the same user. This can cause dependency issues such as failure of creating the spacerole if the user is not an org member anymore or failure of deleting the org role if the user is still assigned a spacerole. Switching to a dedicated user for each resource type mitigates these problems


SatabdiG (approved):
lgtm

PR #271, opened 2026-04-30
feat(app): include route URLs in App status observation

### LINKED ISSUE #190: [FEATURE] `App` status should include the app URL
**Context**
Once the `App` is ready,  app URL should be reported as part of the App status.

**Scope**
URL is computed from `routes`. Currently the app status does include the[ app manifest](https://v3-apidocs.cloudfoundry.org/version/3.209.0/#generate-a-manifest-for-an-app). This should includes a list of `routes`.  

**Out of scope**
If needed, describe what parts of a feature can be ignored for now.

_Hint_: Smaller feature wishes tend to be implemented and shipped quicker!

**Technical Steps** 
Currently the app status does include the[ app manifest](https://v3-apidocs.cloudfoundry.org/version/3.209.0/#generate-a-manifest-for-an-app). This should includes a list of `routes`. 
Todo:  Validate `status.appManifest` contain valid `routes`.

gergely-szabo-sap:
Surface the routes mapped to a Cloud Foundry application in status.atProvider.routes, so users can discover route URLs directly from the Kubernetes resource without querying the CF API.

Closes #190

## Changes

**API** (status-only, backward compatible)

- Added AppRouteObservation struct with fields URL, Host, Path, Protocol, Port
- Added Routes []AppRouteObservation to AppObservation (all fields omitempty)
- Regenerated deepcopy and CRD YAML via controller-gen

**Controller**

- Added RouteFetcher interface, embedded in app.Client
- Implemented FetchRoutes(ctx, appGUID) — calls client.Routes.ListForAppAll
- Called in Observe() as a non-fatal step: on success, fresh routes are written to status; on error, previously observed routes are preserved so transient CF API failures don't erase known data

**Tests**

- Unit tests for FetchRoutes (success, empty result, nil fetcher, API error)
- Controller tests for RoutesPopulated and RouteFetchErrorNonFatal (with route status assertion via cmp.Diff)
- E2E assess steps verifying status.atProvider.routes is populated for both e2e-app and e2e-app-2

**Upgrade safety**

This is an additive, optional, status-only change. No upgrade test is needed:

- Old controller + new CRD: old controller ignores the new field, writes nil routes — benign
- New controller + old CRD: Kubernetes prunes unknown status fields — benign
- No existing data at risk: Routes is purely observational, derived from the CF API on each reconcile


apis/resources/v1alpha1/app_types.go
```diff
@@ -22,6 +22,33 @@ type AppObservation struct {
...
+
+	// The list of routes currently mapped to the application.
+	// +kubebuilder:validation:Optional
+	Routes []AppRouteObservation `json:"routes,omitempty"`
+}
+
+// AppRouteObservation represents an observed route for the application.
+type AppRouteObservation struct {
+	// The full URL of the route (e.g. myapp.apps.example.com).
+	// +kubebuilder:validation:Optional
```
SatabdiG: AppRouteObservation is under AppStatus, I thought Validation markers on Status-only structs have no effect, do we really need them?

internal/controller/app/controller.go
```diff
@@ -139,13 +139,26 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 	appManifest, err := c.client.GenerateManifest(ctx, res.GUID)
 	if err == nil {
 		cr.Status.AtProvider.AppManifest = appManifest
 	}
 
+	// Fetch routes for the application. On success, update the status with
+	// the fresh data; on error, restore the previously observed routes so
+	// that a transient CF API failure does not erase known route information.
+	if routes, err := c.client.FetchRoutes(ctx, res.GUID); err == nil {
+		cr.Status.AtProvider.Routes = routes
```
SatabdiG, line 158: nitpick: Do you think a log debug or warning event would make sense here,  in case of a persistent CF Routes API outage or permission, the underlying issue would be invisible.

SatabdiG (approved):
Very clean, some minor comments from my side but otherwise looks good

internal/controller/app/controller.go
gergely-szabo-sap: Good observation, klog.Warning added

PR #272, opened 2026-05-06
fix: implement retry logic for ServiceCredentialBinding creation attempts

SatabdiG:
## Problem
When a `ServiceCredentialBinding creation` fails (e.g. due to an invalid broker parameter or a feature not enabled in a landscape), the controller enters an infinite reconciliation loop, repeatedly attempting to create new bindings in CF. This causes the CF org to accumulate a large number of failed bindings, eventually making the org unresponsive and preventing deletion of the affected service instance.
The root cause was in `HandleObservationState`: when `LastOperation.State=failed` and `LastOperation.Type=create`, it returned `ResourceExists: false`, which signals crossplane-runtime to call `Create()` again immediately on the next reconcile cycle.
Fixes [Issue](https://github.tools.sap/openmcp/support/issues/586)

## Solution
Introduce a circuit breaker using a crossplane-provider-cloudfoundry/create-attempts annotation on the CR that tracks the number of creation attempts.

Related:
Ticket:  [Issue](https://github.tools.sap/openmcp/support/issues/586)
CF org impact: unresponsive org in JP31 landscape, v0.3.3

### DIFF
--- a/internal/controller/servicecredentialbinding/controller.go
+++ b/internal/controller/servicecredentialbinding/controller.go
@@ -36,7 +36,6 @@ const (
 	errNewClient             = "cannot create a client for " + externalSystem + ": %w"
 	errWrongCRType           = "managed resource is not a " + resourceType
 	errGet                   = "cannot get " + resourceType + " in " + externalSystem + ": %w"
-	errFind                  = "cannot find " + resourceType + " in " + externalSystem
 	errCreate                = "cannot create " + resourceType + " in " + externalSystem + ": %w"
 	errUpdate                = "cannot update " + resourceType + " in " + externalSystem + ": %w"
 	errDelete                = "cannot delete " + resourceType + " in " + externalSystem + ": %w"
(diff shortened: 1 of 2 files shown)


internal/controller/servicecredentialbinding/controller.go
```diff
@@ -272,10 +289,11 @@ func (c *external) HandleObservationState(serviceBinding *cfresource.ServiceCred
 	case v1alpha1.LastOperationFailed:
 		cr.SetConditions(xpv1.Unavailable().WithMessage(serviceBinding.LastOperation.Description))
 		return managed.ExternalObservation{
-			ResourceExists:   serviceBinding.LastOperation.Type != v1alpha1.LastOperationCreate, // set to false when the last operation is create, hence the reconciler will retry create
+			ResourceExists:   true,
 			ResourceUpToDate: serviceBinding.LastOperation.Type != v1alpha1.LastOperationUpdate, // set to false when the last operation is update, hence the reconciler will retry update
 		}, nil
 	case v1alpha1.LastOperationSucceeded:
+		resetCreateAttempts(cr)
```
Kukkerem: Doesn't it mutate only the in-memory cr? Shouldn't we update the object? I think in this `true,true` path, we are only calling  Status().Update().
```
resetCreateAttempts(cr)
if err := c.kube.Update(ctx, cr); err != nil {
    return managed.ExternalObservation{}, fmt.Errorf("cannot persist create attempt reset: %w", err)
}
```
Otherwise, the counter survives a successful create, and a later rotation can trip the circuit breaker prematurely.

SatabdiG: Yes! Many thanks for catching this, adding  an explicit kube.Update(ctx, cr) call right after resetCreateAttempts(cr) in the LastOperationSucceeded case, which writes the annotation removal back to the Kubernetes API server directly, not relying on crossplane-runtime to do it.

mpechkurov (approved):

Kukkerem (approved):
LGTM 🔥

mpechkurov (approved):

PR #277, opened 2026-05-07
docs: avoid service instance default parameters

### LINKED ISSUE #276: [FEATURE] Document service instance default parameter handling is inconsistent
**Context**
When creating a service instance via BTP cockpit, some service offerings have default parameters. Users might reasonably expect that the same default parameters apply when creating a service instance through the CF API (and this provider). This is not consistently the case.

**Scope**
Document in the CRD that default parameter handling varies between service offerings, and that we recommend specifying all of them in the MR.

**Additional context**
The [CF docs](https://v3-apidocs.cloudfoundry.org/version/3.207.0/index.html#create-a-service-instance) and [OSB spec for provisioning](https://github.com/cloudfoundry/servicebroker/blob/master/spec.md#provisioning) do not mention defaults at all. Only the [OSB spec for instance updates](https://github.com/cloudfoundry/servicebroker/blob/master/spec.md#updating-a-service-instance) make a recommendation.

An incomplete list of behaviours:
- `xsuaa` has no default parameters
- `hana` has default parameters in the cockpit only
- `jobscheduler` has default parameters in cockpit and CF API
- `redis-cache` has two _different_ sets of default parameters

johanngoltz:
closes #276

### DIFF
--- a/apis/resources/v1alpha1/serviceinstance_types.go
+++ b/apis/resources/v1alpha1/serviceinstance_types.go
@@ -70,9 +70,9 @@ type Managed struct {
 
 	// (Attributes) Configuration parameters for the managed service instance, supplied as a K8S runtime.RawExtension object.
 	// 
-	// A service offering may set arbitrary default parameters.
-	// These may change at any time for new service instances.
-	// It is recommended to explicitly specify all parameters.
+	// Default parameters applied by the BTP cockpit may differ from those
+	// applied via the CF API. It is recommended to explicitly specify all
+	// parameters to avoid unexpected behavior.
     //
 	// The `parameters` field is NOT secret or secured in any way and should NEVER be used to hold sensitive information.
 	// To set parameters that contain secret information, you should ALWAYS store that information in a Secret and use the `paramsSecretRef` field.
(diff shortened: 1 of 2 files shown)


johanngoltz:
Not sure how we can best get the point across that the default parameters may not match what is proposed in the BTP cockpit: Can we just reference BTP, even if this is the provider for _CF_?

apis/resources/v1alpha1/serviceinstance_types.go
```diff
@@ -69,6 +69,11 @@ type Managed struct {
 	ServicePlan *ServicePlanParameters `json:"servicePlan,omitempty"`
 
 	// (Attributes) Configuration parameters for the managed service instance, supplied as a K8S runtime.RawExtension object.
+	// 
+	// A service offering may set arbitrary default parameters.
+	// These may change at any time for new service instances.
+	// It is recommended to explicitly specify all parameters.
+    //
```
SatabdiG, line 76: I think it's fine to reference BTP Cockpit here, CF Provider users likely afre familar with btp cockpit and maybe suprised when the behavior differs, we could use wording like: 

```
// Default parameters applied by the BTP cockpit may differ from those
// applied via the CF API. It is recommended to explicitly specify all
// parameters to avoid unexpected behavior.

```

SatabdiG (commented):
Thank you taking this up. I think it beneficial to mention the cockpit here and would add clarity to the provider user

apis/resources/v1alpha1/serviceinstance_types.go
johanngoltz: Glad to hear I was overly cautious there! 1812dac72fdec523168e27cbcb5ca411bbaad868

SatabdiG (approved):

PR #284, opened 2026-05-11
feat: support environment variables for App (#188)

### LINKED ISSUE #188: [FEATURE] Support environment variable for `App`
**Context**
The `App` spec includes a field [`Environment`](https://github.com/SAP/crossplane-provider-cloudfoundry/blob/main/apis/resources/v1alpha1/app_types.go#L94) to let user specify optional environment variables. This is currently not processed by the controller. 

**Scope**
When user provides `spec.environment`, the controller should process that, both for creation, and for drift detection, and for update

**Technical Steps** 

- app client to map environment variables in [`newCreateOption`](https://github.com/SAP/crossplane-provider-cloudfoundry/blob/main/internal/clients/app/app.go#L233)
- app controller should [observe environment variables](https://v3-apidocs.cloudfoundry.org/version/3.209.0/#get-environment-variables-for-an-app) and perform drift detection.
- app controller should [update environment variables](https://v3-apidocs.cloudfoundry.org/version/3.209.0/#update-environment-variables-for-an-app) in case of drift. 
**Workarounds & Alternatives**

**Additional context**

1122Louis:
Closes #188

Changes
Architecture
Replaced UpdateAndPush with direct CF Environment Variable API (PATCH /v3/apps/:guid/environment_variables) for env var updates: no push, no restage, no unnecessary restart.

Bug fixes
Fixed nil deletion bug: updateEnvVars now calls GET /v3/apps/:guid/environment_variables first, then sends nil for any var present in CF but absent from spec , so removed vars are actually deleted.
Fixed resulting infinite update loop: was caused by removed vars never being cleared from CF, keeping drift alive.
Fixed double manifest parse: DetectChanges now calls getAppManifest once and shares the result across all checks.

Refactoring
Separated docker_image and environment handling in the controller: docker credentials are only fetched when the docker image actually changes.
Added GetEnvironmentVariables and SetEnvironmentVariables to AppClient interface and fake mock.

Tests
Added 5 unit tests covering env var drift detection (app_test.go):
Same in spec and CF → no drift
Different in spec vs CF → drift detected
Set in spec, missing in CF → drift detected
Removed from spec, still in CF → drift detected (the nil deletion bug)
Nil in both → no drift
Added 3 controller TestUpdate cases (controller_test.go):
EnvVarAdded — new var in spec, CF has none → SetEnvironmentVariables called correctly
EnvVarDeleted — var removed from spec, CF still has it → sent as nil to CF
EnvVarGetFails — GetEnvironmentVariables errors → update returns wrapped error


internal/clients/app/app.go
```diff
@@ -185,6 +155,22 @@ func (cd *ChangeDetection) HasField(field string) bool {
 	return ok
 }
 
+// envVarsChanged returns true if the spec environment variables differ from the current manifest.
+func envVarsChanged(spec v1alpha1.AppParameters, status v1alpha1.AppObservation) (bool, error) {
+	if spec.Environment == nil || spec.Environment.Raw == nil {
+		return false, nil
```
SatabdiG: If the spec had env vars before and now they're removed (set to nil), this returns false — meaning drift is never detected for deletion of env vars. An app that had FOO=bar in CF will never get cleaned up if the user removes the field from the spec.

internal/clients/app/app.go
```diff
@@ -204,6 +190,15 @@ func DetectChanges(spec v1alpha1.AppParameters, status v1alpha1.AppObservation)
 		}
 	}
 
+	// Check if environment variables changed
+	envChanged, err := envVarsChanged(spec, status)
```
SatabdiG: In DetectChanges, the docker image check calls getAppManifest(...) and then envVarsChanged calls it again internally. That's two YAML unmarshals of the same string per reconcile loop. We should refactor this to one. This also makes it harder to test

internal/controller/app/controller.go
```diff
@@ -223,7 +208,7 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
 		return managed.ExternalUpdate{}, errors.Wrap(err, errUpdateResource+": Failed to detect changes")
 	}
 
-	if changes.HasField("docker_image") {
+	if changes.HasField("docker_image") || changes.HasField("environment") {
 		dockerCredentials, err := getDockerCredential(ctx, c.kube, cr.Spec.ForProvider)
```
SatabdiG: This would get called for every environment change even if no docker image is created. It would call getDockerCredential which accesses Kubernetes secrets for no reason.

SatabdiG (changes requested):
We should definitely add test file. 

As far as I am aware there are two ways CF handles env vars
1. The Manifest/Push path - A sledgehammer of sorts, applies a full manifest, res-stages, restarts the app. 
2. The Environment Variable API — CF has a dedicated [PUT /v3/apps/:guid/environment_variables](https://v3-apidocs.cloudfoundry.org/#update-environment-variables-for-an-app) endpoint. This is more of a scalpel and doesn't require a full restage of the app.

My core concern and question is that we are reusing the existing pattern for docker image changed -> push, does env variable need a push? It likely does not if we go via the API. Also `push` are expensive even more for crossplane reconcilers than the plain `cf push`. 

Maybe you could help me through your reasoning here.

internal/clients/app/push.go
```diff
@@ -112,9 +112,14 @@ func newManifestFromSpec(forProvider v1alpha1.AppParameters, dockerCredentials *
...
 
-	if forProvider.LogRateLimitPerSecond != nil {
-		manifest.LogRateLimitPerSecond = *forProvider.LogRateLimitPerSecond
+	if forProvider.Environment != nil && forProvider.Environment.Raw != nil {
+		var envVars map[string]string
+		if err := json.Unmarshal(forProvider.Environment.Raw, &envVars); err != nil {
+			return nil, errors.Wrap(err, "failed to unmarshal environment variables")
+		}
+		manifest.Env = envVars
 	}
```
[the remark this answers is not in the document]
SatabdiG: Copilot is right here, : the LogRateLimitPerSecond mapping was accidentally dropped when adding the env var block, we need to restore it back.

internal/controller/app/controller.go
```diff
@@ -242,6 +246,45 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
+	currentVars, err := c.client.GetEnvironmentVariables(ctx, guid)
+	if err != nil {
+		return errors.Wrap(err, errUpdateResource)
+	}
+	for k := range currentVars {
+		if _, exists := envVars[k]; !exists {
+			envVars[k] = nil
+		}
+	}
+	_, err = c.client.SetEnvironmentVariables(ctx, guid, envVars)
```
SatabdiG, line 279: I think according to the CF API docs: after `PATCH /v3/apps/:guid/environment_variables`, the app needs to be restarted for the new vars to take effect. The controller currently calls SetEnvironmentVariables but never restarts the app. The app will show Synced: True but the running process will still have the old environment.
We would need to call Stop + Start (or a restart endpoint if available in go-cfclient) after a successful SetEnvironmentVariables.

internal/controller/app/controller_test.go
```diff
@@ -571,6 +590,88 @@ func TestUpdate(t *testing.T) {
...
+					withStatus(guid, "STARTED"),
+					withObservedName(name),
+					withAppManifest("applications:\n- name: "+name+"\n  env:\n    ANOTHER_VAR: world")),
+				obs: managed.ExternalUpdate{},
+				err: nil,
+			},
+			service: func() *fake.MockApp {
+				m := &fake.MockApp{}
+				world := "world"
+				m.On("GetEnvironmentVariables", guid).Return(map[string]*string{"ANOTHER_VAR": &world}, nil)
```
SatabdiG, line 638: I don't think `SetEnvironmentVariables` is actually called; testify's mock.On sets up return values but doesn't assert the call happened unless you call m.AssertExpectations(t) at the end. None of the TestUpdate cases call m.AssertExpectations(t). This would mean that the tests verify the return value of Update but not that the correct CF API calls were made. So the tests pass even if SetEnvironmentVariables is never called.

Potential fix: add `tc.service().AssertExpectations(t)` at the end of the test loop, or restructure to hold a reference to the mock.

internal/controller/app/controller_test.go
```diff
@@ -571,6 +590,88 @@ func TestUpdate(t *testing.T) {
 				return m
 			},
 		},
+
+		"EnvVarAdded": {
```
SatabdiG, line 586: Has the same assert expectation gap stated above.

SatabdiG (commented):
good work on the restructure.

internal/controller/app/controller.go
```diff
@@ -242,6 +246,53 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
+	if err != nil {
+		return errors.Wrap(err, errUpdateResource)
+	}
+	// Restart the app so the updated environment takes effect in the running process.
+	if _, err = c.client.Stop(ctx, guid); err != nil {
+		return errors.Wrap(err, errUpdateResource)
+	}
+	_, err = c.client.Start(ctx, guid)
+	return errors.Wrap(err, errUpdateResource)
```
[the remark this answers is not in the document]
SatabdiG: So I think we have two bugs now, copilot is right

1. Say the app is stopped by the user and the env variables get updated, the code now will start the previous stopped app. We need to check the state, If the app was STOPPED, just set the env vars and return — they'll take effect whenever it's next started.
2. updateDockerImage calls UpdateAndPush, which does a full push — CF restages and restarts the app as part of the push operation. Then updateEnvVars calls SetEnvironmentVariables and then Stop/Start again. So if both fields changed simultaneously, the app gets restarted twice — once from the push, once from the env var update. That's unnecessary downtime on top of the downtime already incurred by the push.
We could when docker_image also changed, skip the restart in updateEnvVars because the push already handles it. 

I also think this does reveal a test gap, the current tests for EnvVarAdded and EnvVarDeleted both have the app in STARTED state and neither tests the STOPPED case, nor the combined docker+env change case.

apis/resources/v1alpha1/app_types.go
```diff
@@ -109,7 +109,7 @@ type AppParameters struct {
 	// +kubebuilder:validation:Optional
 	// Sidecars []SidecarConfiguration `json:"sidecars,omitempty"`
 
-	// (NOT SUPPORTED YET) A key-value mapping of environment variables to be used for the app when running
+	// A key-value mapping of environment variables to be used for the app when running
 	// +kubebuilder:validation:Optional
 	Environment *runtime.RawExtension `json:"environment,omitempty"`
```
SatabdiG: We should probably have a kubebuiler validation here, we know it's map[string]string.

internal/controller/app/controller.go
```diff
@@ -242,6 +246,53 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
+	// Build desired env vars from spec
+	envVars := map[string]*string{}
+	if cr.Spec.ForProvider.Environment != nil && cr.Spec.ForProvider.Environment.Raw != nil {
+		raw := map[string]string{}
+		if err := json.Unmarshal(cr.Spec.ForProvider.Environment.Raw, &raw); err != nil {
+			return errors.Wrap(err, errUpdateResource)
+		}
+		for k, v := range raw {
+			v := v
+			envVars[k] = &v
```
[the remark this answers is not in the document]
SatabdiG: A kubebuilder validation on the types would handle this.

internal/clients/app/push.go
```diff
@@ -115,6 +115,15 @@ func newManifestFromSpec(forProvider v1alpha1.AppParameters, dockerCredentials *
 	if forProvider.LogRateLimitPerSecond != nil {
 		manifest.LogRateLimitPerSecond = *forProvider.LogRateLimitPerSecond
 	}
+
+	if forProvider.Environment != nil && forProvider.Environment.Raw != nil {
+		var envVars map[string]string
+		if err := json.Unmarshal(forProvider.Environment.Raw, &envVars); err != nil {
+			return nil, errors.Wrap(err, "failed to unmarshal environment variables")
+		}
```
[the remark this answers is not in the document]
SatabdiG: A kubebuilder validation would fix this.

internal/clients/app/app.go
```diff
@@ -185,25 +195,67 @@ func (cd *ChangeDetection) HasField(field string) bool {
...
 	}
 
+	// Check if environment variables changed
+	envChanged, err := envVarsChanged(spec, appManifest)
+	if err != nil {
+		return nil, err
+	}
+	if envChanged {
+		changes.ChangedFields["environment"] = struct{}{}
+	}
```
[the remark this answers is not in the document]
SatabdiG: Copilot is pointing at the symptom. The actual cause would be in the controller. DetectChanges has no way to distinguish a genuinely empty manifest from a stale "" written by a failed GenerateManifest call — the fix must be in Observe, not in DetectChanges.

internal/clients/app/app.go
```diff
@@ -185,25 +195,67 @@ func (cd *ChangeDetection) HasField(field string) bool {
...
+	if spec.Lifecycle == "docker" && spec.Docker != nil && appManifest.Docker != nil {
+		if spec.Docker.Image != appManifest.Docker.Image {
+			changes.ChangedFields["docker_image"] = struct{}{}
 		}
 	}
 
+	// Check if environment variables changed
+	envChanged, err := envVarsChanged(spec, appManifest)
+	if err != nil {
+		return nil, err
```
SatabdiG: In `Observe`, `GenerateObservation` always resets AtProvider (including AppManifest to "") before `GenerateManifest` is called. If `GenerateManifest` fails transiently, `AppManifest` is left as "" and the error is silently swallowed:

`IsUpToDate` is then called in the same `Observe` pass against this now-empty manifest. `DetectChanges` treats a missing manifest as a zero-value AppManifest{} — so any env vars present in spec are compared against an empty map, producing false drift. Observe returns ResourceUpToDate: false, the reconciler persists the "" manifest to status, and immediately calls Update in the same cycle — triggering a spurious SetEnvironmentVariables and a Stop/Start restart against a running app for no reason.

The routes handling in the same function already has the right pattern, we can apply it here.

SatabdiG (changes requested):
Nice work but I think it introduced some new bugs

1122Louis:
Thanks for the detailed review. Here's what I addressed:

1. STOPPED app being started / double restart on docker+env change
Fixed in updateEnvVars, it now takes a dockerAlsoChanged bool parameter. If the app is STOPPED or a docker push already happened (which CF uses to restart), the Stop/Start is skipped.

2. Missing test cases
Added EnvVarUpdated_AppStopped (verifies no Stop/Start when app is STOPPED) and EnvVarAndDockerChanged (verifies no Stop/Start when docker image also changed).

3. kubebuilder validation for environment field
Changed the type from *runtime.RawExtension to map[string]string. This makes the type itself the validation, controller-gen generates additionalProperties: type: string in the CRD schema, so Kubernetes rejects non-string values at the API level. No explicit marker needed, and it also removes all the json.Unmarshal boilerplate from the controller and push.go.

4. AppManifest cleared to "" on transient GenerateManifest failure
Fixed in Observe, it saves prevAppManifest before resetting AtProvider, and restores it if GenerateManifest fails, so DetectChanges is never comparing against a stale empty manifest.

5. nil-panic in SetEnvironmentVariables mock
Added the same nil guard that GetEnvironmentVariables already had.

internal/clients/app/app.go
```diff
@@ -185,25 +193,61 @@ func (cd *ChangeDetection) HasField(field string) bool {
...
 		}
-		if appManifest.Docker != nil {
-			if spec.Docker.Image != appManifest.Docker.Image {
-				changes.ChangedFields["docker_image"] = struct{}{}
-			}
+		appManifest = m
+	}
+
+	// Check if Docker image changed
+	if spec.Lifecycle == "docker" && spec.Docker != nil && appManifest.Docker != nil {
```
SatabdiG: I guess I already pointed this out in a previous review. If AppManifest is empty for any reason (new app, first reconcile before manifest is populated), docker image drift is silently skipped. ResourceUpToDate: true is returned and the image change is never actioned.

internal/controller/app/controller_test.go
```diff
@@ -564,10 +578,156 @@ func TestUpdate(t *testing.T) {
...
+				m := &fake.MockApp{}
+				v := "hello"
+				m.On("GetEnvironmentVariables", guid).Return(map[string]*string{}, nil)
+				m.On("SetEnvironmentVariables", guid, map[string]*string{"MY_VAR": &v}).Return(map[string]*string{}, nil)
+				// No Stop/Start expected — app is already stopped
+				return m
+			},
+		},
+
+		"EnvVarAndDockerChanged": {
```
SatabdiG, line 701: the mock sets up m.On("Update", guid) for the UpdateAndPush path, but UpdateAndPush in app.go calls c.AppClient.Update and then c.Push. The Push mock in newMockPush() is set up with m.On("Push"). However newMockPush().GenerateManifest is set up for guid specifically — and it returns a manifest with just the app name, no docker section. So after updateDockerImage runs, the manifest in status still has old-image:v1. So for me, this test isn't realistic e2e

internal/controller/app/controller_test.go
```diff
@@ -564,10 +578,156 @@ func TestUpdate(t *testing.T) {
...
+				err: nil,
+			},
+			service: func() *fake.MockApp {
+				m := &fake.MockApp{}
+				v := "hello"
+				// UpdateAndPush (docker image update) calls AppClient.Update
+				m.On("Update", guid).Return(&fake.NewApp("docker").SetName(name).SetGUID(guid).App, nil)
+				m.On("GetEnvironmentVariables", guid).Return(map[string]*string{}, nil)
+				m.On("SetEnvironmentVariables", guid, map[string]*string{"MY_VAR": &v}).Return(map[string]*string{}, nil)
+				// No Stop/Start expected — docker push already restarted the app
```
SatabdiG, line 737: We should have a test for the partial failure scenario, Stop fails / Start fails tests — if Stop succeeds but Start fails, app is left STOPPED with new env vars applied.

SatabdiG (changes requested):

internal/clients/app/app.go
1122Louis: Fixed — changed the condition to appManifest.Docker == nil || spec.Docker.Image != appManifest.Docker.Image, so missing docker in the manifest is now treated as drift. Also made Observe return an error if GenerateManifest fails, so we never reach DetectChanges with an empty manifest for an existing app.

internal/controller/app/controller_test.go
1122Louis: Replaced the shared newMockPush() with a dedicated push mock that only expects Push (no GenerateManifest expectation). Also added AssertExpectations on the push mock for this case so we actually verify the expected calls happened. EnvVarAndDockerChanged test gave it its own push mock that only expects Push (no GenerateManifest), with AssertExpectations to catch unexpected calls. Gave it its own push mock that only expects Push (no GenerateManifest), with AssertExpectations to catch unexpected calls.

internal/controller/app/controller_test.go
1122Louis: Stop/Start failure tests — added EnvVarUpdate_StopFails and EnvVarUpdate_StartFails. 
EnvVarUpdate_StopFails — Stop returns error → wrapped error propagated
EnvVarUpdate_StartFails — Stop succeeds, Start returns error → wrapped error propagated

Both verify the error bubbles up wrapped in [errUpdateResource]

internal/clients/app/app.go
```diff
@@ -87,6 +90,11 @@ func (c *Client) CreateAndPush(ctx context.Context, spec v1alpha1.AppParameters,
 	return c.Push(ctx, application, manifest, nil)
 }
 
+// GetEnvironmentVariables gets the environment variables for an app.
+func (c *Client) GetEnvironmentVariables(ctx context.Context, guid string) (map[string]*string, error) {
```
SatabdiG: Client embeds AppClient, so this method is already promoted automatically. The explicit wrapper is dead code — calling c.AppClient.GetEnvironmentVariables directly would work fine without it, and it's the only method on AppClient that got this treatment.

We should remove this code unless there is a good reason for it. Then we add an comment

SatabdiG (commented):
Very good work, this is almost there. I have just one comment.

SatabdiG (approved):
lgtm. Thanks for sticking with this 🥇

PR #285, opened 2026-05-19
feat: add workflow to skip checks for documentation-only pull requests

SatabdiG:
## Summary

Adds a pr-skip.yaml workflow to unblock doc-only PRs. When a PR only touches .md files, the main pr-test.yaml workflow is skipped due to paths-ignore, leaving required checks stuck in "waiting" state. 

This workflow satisfies those required checks by running license compliance checks (go-licenses, reuse) and posting skip stubs for unit tests, reviewable diff, and e2e.


.github/workflows/pr-skip.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+    paths:
+      - '*.md'
```
Kukkerem: I think it only matches on the root level. https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

**.md would match any .md files.

Kukkerem: Also it's probably the same in pr-test.yaml

SatabdiG: Yes absolutely correct on both counts! thanks for the catch

schlingemann:
I'm not an expert about the glob syntax, but here's a few things Claude found :)

### `run-e2e-test` stub probably doesn't satisfy the required check

The real `run-e2e-test` lives in `pr-e2e.yaml` and runs on `pull_request_target`. This new workflow only triggers on `pull_request`, so the stub job won't match the required check from `pr-e2e.yaml` — required checks are matched per workflow run.

If e2e is a required check, `pr-e2e.yaml` needs the same treatment (`paths-ignore: '**.md'` + stub on `pull_request_target`). If it isn't required, just drop the stub here.

### `pr-e2e.yaml` still uses `*.md`

You're fixing the glob in `pr-test.yaml` (`*.md` → `**.md`), but `pr-e2e.yaml` has the same bug and isn't touched. A doc-only PR under `docs/` will now correctly skip `pr-test.yaml` but still trigger e2e. Worth fixing in the same PR.

### Mixed PRs will produce duplicate check runs

`paths`/`paths-ignore` are file-level, not "only these files". A PR touching `README.md` + any `.go` file triggers both workflows, so you'd get two `run-unit-test` checks (one real, one stub). Probably fine for branch protection but noisy — worth verifying on a mixed PR.

### Nit

Header comment says "triggered for pull requests that only modify markdown files" — actually triggers whenever *any* markdown file is touched. Minor wording fix.

SatabdiG:
> I'm not an expert about the glob syntax, but here's a few things Claude found :)
> 
> ### `run-e2e-test` stub probably doesn't satisfy the required check
> The real `run-e2e-test` lives in `pr-e2e.yaml` and runs on `pull_request_target`. This new workflow only triggers on `pull_request`, so the stub job won't match the required check from `pr-e2e.yaml` — required checks are matched per workflow run.
> 
> If e2e is a required check, `pr-e2e.yaml` needs the same treatment (`paths-ignore: '**.md'` + stub on `pull_request_target`). If it isn't required, just drop the stub here.
> 
> ### `pr-e2e.yaml` still uses `*.md`
> You're fixing the glob in `pr-test.yaml` (`*.md` → `**.md`), but `pr-e2e.yaml` has the same bug and isn't touched. A doc-only PR under `docs/` will now correctly skip `pr-test.yaml` but still trigger e2e. Worth fixing in the same PR.
> 
> ### Mixed PRs will produce duplicate check runs
> `paths`/`paths-ignore` are file-level, not "only these files". A PR touching `README.md` + any `.go` file triggers both workflows, so you'd get two `run-unit-test` checks (one real, one stub). Probably fine for branch protection but noisy — worth verifying on a mixed PR.
> 
> ### Nit
> Header comment says "triggered for pull requests that only modify markdown files" — actually triggers whenever _any_ markdown file is touched. Minor wording fix.

Good catch on the e2e tests by claude , since e2e tests are unnecessary for doc-only changes, I've removed run-e2e-test from the required checks in the branch ruleset instead. This avoids wasting CI resources on markdown PRs while keeping the gate for code changes.

The rest done

Kukkerem (approved):
LGTM

PR #286, opened 2026-05-20
App External-Name Handling

### LINKED ISSUE #238: [TASK] `External Name` (App): Ensure ADR Compliance
**Part of Epic: #236** 

## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

lennardrother:
Closes: #238

1. Implemented correct external-name handling in Observe, Create and Delete functions
2. Adapted unit tests for Observe, Create and Delete functions
3. Created E2E import test
4. Created upgrade test for external-name behavior
5. Added documentation


internal/controller/app/controller.go
```diff
@@ -214,8 +227,10 @@ func (c *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
 	}
 
 	guid := meta.GetExternalName(cr)
-	if _, err := uuid.Parse(guid); err != nil {
-		return managed.ExternalUpdate{}, errors.New(errUpdateResource + ": No valid GUID found for the App")
+	// Observe validates GUID format before marking ResourceUpToDate:false, so
+	// Update is only called with a valid GUID.
+	if guid == "" {
```
SatabdiG: nitpick: maybe we can inline this `meta.GetExternalName(cr) == ""`

internal/controller/app/controller_test.go
```diff
@@ -198,15 +229,11 @@ func TestObserve(t *testing.T) {
 			},
 			want: want{
 				mg:  newApp("docker", withExternalName(guid), withSpace(spaceGUID)),
-				obs: managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: true, ResourceLateInitialized: true},
+				obs: managed.ExternalObservation{ResourceExists: true, ResourceLateInitialized: true},
```
SatabdiG, line 233: I saw that the `Should adopt` case correctly sets `want.mg` with `withExternalName(guid)` to verify that `resolveExternalName` calls `meta.SetExternalName` — but unlike `TestCreate` and `TestDelete`, `TestObserve` never diffs tc.want.mg against tc.args.mg after the call. The assertion is silently a no-op.

You could remove `meta.SetExternalName(cr, res.GUID)` from resolveExternalName entirely and this test would still pass.

Maybe we can add the same diff block that TestCreate already has:

test/upgrade/app_external_name_upgrade_test.go
```diff
@@ -0,0 +1,103 @@
...
+				}
+
+				app := &v1alpha1.App{}
+				if err = r.Get(ctx, appName, cfg.Namespace(), app); err != nil {
+					t.Fatalf("Failed to get App resource: %v", err)
+				}
+
+				externalName := xpmeta.GetExternalName(app)
+				if !test.UUIDRegex.MatchString(externalName) {
+					t.Fatalf("External name %q does not match expected UUID format after upgrade", externalName)
```
SatabdiG: I am wondering if this is necessary. Later on we do a stability check 

``` 
if externalName != preUpgradeExternalName {
    t.Fatalf("External name changed during upgrade: ...")
}
```

If that passes, it means `externalName == preUpgradeExternalName`. But `preUpgradeExternalName` was already validated as a UUID in this assessment. So this assessment can never fail independently of the stability check.
The only scenario in which this assessment would catch something that the stability check misses is if the external-name changed to a different non-UUID value after the upgrade — but that's already caught by the following stability check.

SatabdiG (commented):
Great work, some minor comments from my side.

internal/controller/app/controller.go
lennardrother: Done: https://github.com/SAP/crossplane-provider-cloudfoundry/pull/286/commits/ac461fc1e8bcdf3520eb92f3a2c25a644eae2ca9

internal/controller/app/controller_test.go
lennardrother: https://github.com/SAP/crossplane-provider-cloudfoundry/pull/286/commits/6f40d72b6dd76cbf41864721df2cd849472afda6 uses a full diff ignoring `CreatedAt` and `UpdatedAt`. Is this fine?

test/upgrade/app_external_name_upgrade_test.go
lennardrother: You are right, fixed with https://github.com/SAP/crossplane-provider-cloudfoundry/pull/286/commits/1b9c1a1787f39b75f2029e53c30443b4a79e3195

SatabdiG (approved):
🔥

PR #290, opened 2026-05-21
[281#4] fix(exporter): strip Crossplane default labels from exported CF resources

mpechkurov:
## What
- New `cmd/exporter/cf/metadata/metadata.go` with `StripDefaultLabels` helper
- Strip `crossplane-kind`, `crossplane-name`, `crossplane-providerconfig` from exported CR manifests
- Fix import grouping in exporter convert files (goimports convention)

## Why
Default labels are infrastructure metadata computed by the controller from the CR identity. They are semantically incorrect for a new/exported CR (different name, different ProviderConfig) and the controller will recompute them on the next reconcile.

## Part of
Split from #281 — PR 4 of 6. Depends on [281#1] only (can be merged independently of [281#2] and [281#3]).

**Merge order:** [281#1] → [281#2] → [281#3] → [281#4] → [281#5] → [281#6]

### DIFF
--- a/cmd/exporter/cf/app/convert.go
+++ b/cmd/exporter/cf/app/convert.go
@@ -7,9 +7,9 @@ import (
 	"log/slog"
 
 	"github.com/SAP/crossplane-provider-cloudfoundry/apis/resources/v1alpha1"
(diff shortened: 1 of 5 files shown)


SatabdiG:
Closing as https://github.com/SAP/crossplane-provider-cloudfoundry/pull/281 received review comments.

PR #293, opened 2026-05-21
ci: parallelize e2e tests with BUILD_ID-scoped resources

### LINKED ISSUE #274: [FEATURE] Enable parallel execution of e2e tests
**Context**
Currently, the e2e tests run sequentially and share mutable Cloud Foundry resources (orgs, spaces, service instances, routes, etc.) by using hardcoded names. This prevents tests from being executed in parallel, which significantly increases CI feedback time, especially as the test suite grows.

**Goal**
Enable e2e tests to run in parallel by ensuring each test run uses uniquely scoped resource names derived from a \`BUILD_ID\`, eliminating naming conflicts between concurrent test executions.

**Out of scope**
- Local parallelization (not required; developers can run individual test suites)
- Changes to test logic or coverage

**Acceptance Criteria**
- [ ] GitHub Actions matrix runs e2e test suites in parallel
- [ ] All existing e2e tests pass after refactoring

Kukkerem:
Closes #274

Each `Test*` function now runs as its own matrix entry in GitHub Actions, so the e2e suite finishes at the time of the slowest test instead of the sum of all tests. Every test-created CF resource name gets a `-$BUILD_ID` suffix (except the role-import usernames - see Trade-offs) so parallel jobs and concurrent workflow runs don't collide on the shared `cf-ci-e2e` org.

**What changed**

- GitHub Actions discovers `Test*` functions via grep and fans them out (`fail-fast: false`)
- CF resource names are BUILD_ID-scoped via `runScopedName()` / `NewID()` - CRS YAMLs are rendered into `.work/rendered-crs/` with `envsubst '$BUILD_ID'`, tracked files stay clean
- `BUILD_ID` is a workflow-level env in CI (`run_id-run_attempt`) - `make test-acceptance` defaults to `HHMMSS` - `cleanup.go` falls back to `0000` when invoked manually without it
- New `cleanup.go` deletes BUILD_ID-scoped resources after the matrix finishes - checks out `github.base_ref` (not the PR head) so PR authors can't inject code that runs with CF credentials
- `concurrency: e2e-cf-ci-e2e` (`cancel-in-progress: false`) queues concurrent workflow runs - orgrole/spacerole import tests use static `sap.ids` usernames that can't be BUILD_ID-scoped without IDS provisioning
- Removed `resetTestOrg` and the sequential-era CF-cleanup helpers from `test_env.go` (they'd race in parallel - `cleanup.go` does the job now)
- Pre-provisioned CF fixtures (`cf-ci-e2e`, `upgrade-test-org`, `*-donotdelete` spaces, `upgrade-test-import-space`, system domains, sap.ids test users) are Observe-only and untouched

**Trade-offs**

- PR fixes to `cleanup.go` only take effect after merge - base-ref checkout is a deliberate security choice (matches the BTP provider)
- Concurrent workflow runs serialize instead of parallelizing - alternative would need a user pool or shadow-user creation in SAP IDS
- Cleanup is best-effort - canceled runs leave resources until the next pick-up

The valid e2e test run:
https://github.com/SAP/crossplane-provider-cloudfoundry/actions/runs/26230189808


.github/workflows/cleanup.go
```diff
@@ -0,0 +1,192 @@
...
+	return client.New(cfg)
+}
+
+func deleteSCB(ctx context.Context, cfClient *client.Client, name, serviceInstanceName string) error {
+	scb, err := cfClient.ServiceCredentialBindings.Single(ctx, &client.ServiceCredentialBindingListOptions{
+		Names:                client.Filter{Values: []string{name}},
+		ServiceInstanceNames: client.Filter{Values: []string{serviceInstanceName}},
+	})
+	if err != nil {
+		return err
```
SatabdiG: I am wondering why we aren't returning `nil` here as with the other cases. Won't this log a spurious error if a test fails before it?

test/e2e/cloudfoundry_spacequota_import_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var (
 	spaceQuotaImportTestK8sResName  = "e2e-test-space-quota-import"
-	spaceQuotaImportTestName        = "e2e-test-space-quota"
+	spaceQuotaImportTestName        = runScopedName("e2e-test-space-quota")
 	spaceQuotaImportTestOrgName     = "upgrade-test-org"
```
SatabdiG, line 18: I thought `cleanup.go` fetches only `cf-ci-e2e` and calls `deleteQuota` under that org's GUID. The `e2e-test-space-quota-<BUILD_ID> `resources from this test live under `upgrade-test-org` and will not be cleaned up?

.github/workflows/e2e_test.yaml
```diff
@@ -28,25 +28,58 @@ on:
       CF_ENVIRONMENT:
```
SatabdiG, line 28: I know you didn;t touch this but since we are touching this file anyway, we hardcode go-version: go-version: '1.25', do you think it makes sense to derive it from go.mod?

SatabdiG (commented):
I really like this, the security stuff especially 🔥 Just some minor comments

.github/workflows/cleanup.go
Kukkerem: Thank you, good catch!

test/e2e/cloudfoundry_spacequota_import_test.go
Kukkerem: Very true, added it to the cleanup.

.github/workflows/e2e_test.yaml
Kukkerem: Changed it!

test/e2e/cloudfoundry_spacequota_import_test.go
Kukkerem: Also, I found an oversight in the SRB cleanup and fixed it, thanks to this comment.

SatabdiG (dismissed):
lgtm! thanks for taking this up, much needed 🦸

SatabdiG (approved):
lgtm

PR #298, opened 2026-05-29
Migrate CF secrets to environment secrets

jn-av:
Adds `environment: pr-e2e-no-approval` to the `upgrade-test` job, which was the only workflow accessing `CF_CREDENTIALS` and `CF_ENVIRONMENT` without an environment gate. All other workflows (`e2e_test.yaml`, `pr-e2e.yaml`, `release.yaml`) already ran within a GitHub environment.


SatabdiG (approved):
thanks for doing this

PR #301, opened 2026-06-02
fix: add environment to e2e-test and cleanup-e2e jobs

jn-av:
`e2e-test` and `cleanup-e2e` jobs were missing `environment:`, causing CF secrets to be empty. `cleanup-e2e` uses `pr-e2e-no-approval` to prevent approval gates from blocking cleanup.


SatabdiG (approved):
looks good

PR #302, opened 2026-06-03
fix: pin Helm version to v3.14.0 in e2e test workflow

SatabdiG:
## Fix flaky Helm install in e2e matrix

Replaces the manual `packages.buildkite.com` apt-based Helm install with `azure/setup-helm@v4`.

### Problem
Concurrent matrix jobs were intermittently failing with a `403 Forbidden` when fetching the Helm GPG key and apt repository from `packages.buildkite.com`. Jobs sharing the same Azure runner IP range were being rate-limited, causing non-deterministic failures unrelated to test logic.

### Fix
Switch to `azure/setup-helm@fe7b79cd5ee1e45176fcad797de68ecaf3ca4814 # v4.2.0` with Helm `v3.14.0` pinned. This pulls from GitHub releases instead of a third-party apt repo, is cache-friendly on GitHub-hosted runners, and is not subject to the rate limiting that caused the flakiness.

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -90,12 +90,9 @@ jobs:
           go-version-file: 'go.mod'
 
       - name: Install Helm
-        run: |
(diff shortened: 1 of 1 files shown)


Kukkerem (approved):
LGTM

PR #307, opened 2026-06-09
ci(e2e): retry controlplane.up on flaky uxp pull

Kukkerem:
The UXP chart pull from `charts.upbound.io` intermittently fails with transient 403/rate-limit errors under e2e matrix load, causing otherwise-healthy shards to fail during control-plane setup.

Adds a `controlplane.up.retry` target that retries `controlplane.up` up to `CONTROLPLANE_UP_ATTEMPTS` (default 5) with inear backoff (`attempt * 15s`), and routes `local-deploy-prebuilt` through it. Single-attempt `controlplane.up` is unchanged for non-matrix use.


SatabdiG (approved):
🔥

PR #309, opened 2026-06-15
ci: shift e2e approval left

n-losse:
issue: a workflow waiting for approval is already blocking the concurrency slot for other PRs 

since matrix jobs should run in parallel, can't use job level concurrency, need to keep it at workflow level

shift approval left outside of the workflow

now, e2e workflow always runs on the no-approval environment, so remove parameters


SatabdiG (approved):
🔥

PR #311, opened 2026-06-16
ServiceInstance External-Name Handling

### LINKED ISSUE #246: [TASK] `External Name` (ServiceInstance): Ensure ADR Compliance
**Part of Epic: #236** 

## Context

## Technical Tasks Checklist

**Acceptance Criteria:**

- [ ] **Migration Logic**:

**Acceptance Criteria:**
- [ ] **Observe() Tests**:

- [ ] **Create() Tests**:

- [ ] **Update() Tests**:

- [ ] **Delete() Tests**:

**Acceptance Criteria:**
- [ ] **Import Flow Test**:

### 5. Documentation

Kukkerem:
Closes: #246 

3. Created E2E Import Test
4. Created Upgrade Test for External-Name Behavior
5. Added documentation

### DIFF
--- a/apis/resources/v1alpha1/serviceinstance_types.go
+++ b/apis/resources/v1alpha1/serviceinstance_types.go
@@ -252,7 +252,7 @@ type ServiceInstanceStatus struct {
 //   - Follows Standard: yes
 //   - Format: ServiceInstance GUID (UUID format)
 //   - How to find:
-//   - UI: In the BTP Cockpit, navigate to your service instance; the GUID is the ID in the URL
+//   - UI: In the BTP Cockpit, open the service instance detail view; the GUID is shown in the "Instance ID" field
 //   - CLI: `cf service <SERVICE_INSTANCE_NAME> --guid`
 //
 // +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
--- a/apis/resources/v1alpha1/serviceinstance_types.go
+++ b/apis/resources/v1alpha1/serviceinstance_types.go
@@ -247,6 +247,14 @@ type ServiceInstanceStatus struct {
 // +kubebuilder:storageversion
 
 // ServiceInstance is the Schema for the ServiceInstances API. Provides a Cloud Foundry resource for managing service instances.
+//
+// External-Name Configuration:
+//   - Follows Standard: yes
+//   - Format: ServiceInstance GUID (UUID format)
+//   - How to find:
+//   - UI: In the BTP Cockpit, navigate to your service instance; the GUID is the ID in the URL
+//   - CLI: `cf service <SERVICE_INSTANCE_NAME> --guid`
+//
 // +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
 // +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
 // +kubebuilder:printcolumn:name="EXTERNAL-NAME",type="string",JSONPath=".metadata.annotations.crossplane\\.io/external-name"
--- a/internal/clients/serviceinstance/serviceinstance.go
+++ b/internal/clients/serviceinstance/serviceinstance.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/cloudfoundry/go-cfclient/v3/client"
 	"github.com/cloudfoundry/go-cfclient/v3/resource"
-	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"k8s.io/utils/ptr"
 
@@ -82,15 +81,6 @@ func NewClient(cf *client.Client) *Client {
 	return &Client{cf.ServiceInstances, cf.Jobs, cf.ServicePlans}
 }
 
-// GetByIDOrSpec retrieves external resource by GUID or by matching CR's ForProvider spec
-func GetByIDOrSpec(ctx context.Context, c *Client, guid string, spec v1alpha1.ServiceInstanceParameters) (*resource.ServiceInstance, error) {
-	if _, err := uuid.Parse(guid); err == nil {
-		return c.Get(ctx, guid)
-	}
-
-	return c.MatchSingle(ctx, spec)
-}
-
 // Get retrieves external resource using GUID
 func (c *Client) Get(ctx context.Context, guid string) (*resource.ServiceInstance, error) {
 	if guid == "" {
@@ -99,8 +89,22 @@ func (c *Client) Get(ctx context.Context, guid string) (*resource.ServiceInstanc
 	return c.ServiceInstance.Get(ctx, guid)
 }
 
+// GetByGUIDOrSpec resolves the external service instance per the external-name ADR: a set
+// external-name (guid) is fetched directly, while an empty one falls back to a spec-based lookup
+// for backward compatibility. The caller is responsible for validating the GUID format.
+func GetByGUIDOrSpec(ctx context.Context, c *Client, guid string, spec v1alpha1.ServiceInstanceParameters) (*resource.ServiceInstance, error) {
+	if guid != "" {
+		return c.Get(ctx, guid)
+	}
+	return c.MatchSingle(ctx, spec)
+}
+
 // MatchSingle retrieves external resource by matching CR's ForProvider spec
 func (c *Client) MatchSingle(ctx context.Context, spec v1alpha1.ServiceInstanceParameters) (*resource.ServiceInstance, error) {
+	// Backward-compat lookup requires at least a name to match on.
+	if spec.Name == nil || *spec.Name == "" {
+		return nil, nil
+	}
 	// if external-name is not set, search by Name and Space
 	opt := client.NewServiceInstanceListOptions()
 	opt.Type = string(spec.Type)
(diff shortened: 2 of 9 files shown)


apis/resources/v1alpha1/serviceinstance_types.go
```diff
@@ -247,6 +247,14 @@ type ServiceInstanceStatus struct {
 // +kubebuilder:storageversion
 
 // ServiceInstance is the Schema for the ServiceInstances API. Provides a Cloud Foundry resource for managing service instances.
+//
+// External-Name Configuration:
+//   - Follows Standard: yes
+//   - Format: ServiceInstance GUID (UUID format)
+//   - How to find:
+//   - UI: In the BTP Cockpit, navigate to your service instance; the GUID is the ID in the URL
```
lennardrother: You can also find the ID in the detail view in the "Instance ID" field, might be easier

internal/controller/serviceinstance/controller_test.go
```diff
@@ -705,6 +692,11 @@ func TestObserve(t *testing.T) {
 			if diff := cmp.Diff(tc.want.obs, obs); diff != "" {
 				t.Errorf("Observe(...): -want, +got:\n%s", diff)
 			}
+			if tc.want.mg != nil {
+				if diff := cmp.Diff(meta.GetExternalName(tc.want.mg), meta.GetExternalName(tc.args.mg)); diff != "" {
+					t.Errorf("Observe(...): want external-name != got external-name:\n%s", diff)
+				}
+			}
```
lennardrother, line 811: Idea: We could do a full diff ignoring the timestamps as for the App: https://github.com/SAP/crossplane-provider-cloudfoundry/blob/8f20f252fb9b52cb437a083d3c2e7d9efc986214/internal/controller/app/controller_test.go#L446-L450

test/upgrade/service_instance_external_name_upgrade_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+				serviceInstance := &v1alpha1.ServiceInstance{}
+
+				err = r.Get(ctx, serviceInstanceName, cfg.Namespace(), serviceInstance)
+				if err != nil {
+					t.Fatalf("Failed to get ServiceInstance resource: %v", err)
+				}
+
+				annotations := serviceInstance.GetAnnotations()
+				externalName, exists := annotations["crossplane.io/external-name"]
```
lennardrother: Should we use meta.GetExternalName() here instead?

apis/resources/v1alpha1/serviceinstance_types.go
SatabdiG: I think there is an indentation off here, `UI` and `CLI` are children of `How to find` right?

Kukkerem: Thank you! Fixed it.

internal/controller/serviceinstance/controller_test.go
Kukkerem: Great idea, fixed!

test/upgrade/service_instance_external_name_upgrade_test.go
Kukkerem: Good insight, fixed.

internal/clients/serviceinstance/serviceinstance.go
```diff
@@ -99,8 +89,22 @@ func (c *Client) Get(ctx context.Context, guid string) (*resource.ServiceInstanc
...
+func GetByGUIDOrSpec(ctx context.Context, c *Client, guid string, spec v1alpha1.ServiceInstanceParameters) (*resource.ServiceInstance, error) {
+	if guid != "" {
+		return c.Get(ctx, guid)
+	}
+	return c.MatchSingle(ctx, spec)
+}
+
 // MatchSingle retrieves external resource by matching CR's ForProvider spec
 func (c *Client) MatchSingle(ctx context.Context, spec v1alpha1.ServiceInstanceParameters) (*resource.ServiceInstance, error) {
+	// Backward-compat lookup requires at least a name to match on.
```
SatabdiG, line 104: I know you didn't touch this part, but reading it together, we might have a bug. 

I see `pollJobComplete` intentionally swallows async timeout errors (comment explains this is by design, deferring to later state observation), and the subsequent `MatchSingle` call returns `(nil, nil)` when no match is found yet — treating "not visible yet" the same as "doesn't exist". 

So `c.serviceinstance.Create` can return `(nil, nil)` for a managed service instance whose creation is still in progress when the poll times out. `err` is nil, so the guard above doesn't catch it, and `r.GUID` panics.

Maybe we can safeguard it here, make external name handling more robust, what do you think?

internal/controller/serviceinstance/controller.go
```diff
@@ -331,8 +339,14 @@ func (c *external) Delete(ctx context.Context, mg resource.Managed) (managed.Ext
 	}
 	cr.SetConditions(xpv1.Deleting())
 
-	if err := c.serviceinstance.Delete(ctx, cr); err != nil {
-		return managed.ExternalDelete{}, errors.New(errDelete)
+	// ADR: the external-name (GUID) identifies the resource to delete; nothing to do if unset.
+	guid := meta.GetExternalName(cr)
+	if guid == "" {
```
SatabdiG, line 358: minor: Should we also validate the guid here? Worse case here is CF-side crash but still.

internal/controller/serviceinstance/controller.go
```diff
@@ -140,19 +140,25 @@ func (c *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		return managed.ExternalObservation{}, errors.New(errWrongCRType)
 	}
```
SatabdiG, line 141: nit: You didn;t touch part of this code: 

```
if _, err := uuid.Parse(guid); err == nil {
		// We have a valid external-name annotation
		return nil
	}
``` 

but should we replace `uuid.Parse(guid)` with `clients.IsValidGUID` for a single source of truth on what counts as a valid external-nam

internal/controller/serviceinstance/controller_test.go
```diff
@@ -187,7 +189,7 @@ func TestObserve(t *testing.T) {
 				mg: serviceInstance("managed", withExternalName(guid), withSpace(spaceGUID), withServicePlan(v1alpha1.ServicePlanParameters{ID: &servicePlan})),
 			},
 			want: want{
-				mg:  serviceInstance("managed", withExternalName(guid)),
+				mg:  serviceInstance("managed", withExternalName(guid), withSpace(spaceGUID), withServicePlan(v1alpha1.ServicePlanParameters{ID: &servicePlan})),
```
SatabdiG, line 192: I see TestObserve, TestCreate, TestCreate case structs declare the kube k8s.Client. 

But the test runner never reads `tc.kube`; it always constructs a hardcoded `&test.MockClient{MockUpdate: test.NewMockUpdateFn(nil)}`. Wouldn't that mean there's currently no way to simulate `c.kube.Update `failing? So in return the Observe's adoption path `(meta.SetExternalName + c.kube.Update)` and `postCreate (c.kube.Update then c.kube.Status().Update)` both have an `errUpdateCR` error path that would never exercised?

test/upgrade/service_instance_external_name_upgrade_test.go
```diff
@@ -0,0 +1,116 @@
...
+				}
+
+				err = v1alpha1.SchemeBuilder.AddToScheme(r.GetScheme())
+				if err != nil {
+					t.Fatalf("Failed to add CloudFoundry scheme: %v", err)
+				}
+
+				serviceInstance := &v1alpha1.ServiceInstance{}
+
+				err = r.Get(ctx, serviceInstanceName, cfg.Namespace(), serviceInstance)
```
SatabdiG, line 59: Should we also use `SchemeBuilder` here same as `PreUpgrade`

SatabdiG (commented):
Good work, Some comments from my side.

apis/resources/v1alpha1/serviceinstance_types.go
Kukkerem: Other resources has the same because gofmt rewrites the comment in case of nesting. We could definitely work on this, but currently gofmt would fail.

internal/clients/serviceinstance/serviceinstance.go
Kukkerem: Very true, fixed with a new UT!

internal/controller/serviceinstance/controller.go
Kukkerem: Good point, fixed as well.

internal/controller/serviceinstance/controller.go
Kukkerem: True, fixed, dropped the now unused uuid.

internal/controller/serviceinstance/controller_test.go
Kukkerem: Good catch, fixed!

test/upgrade/service_instance_external_name_upgrade_test.go
Kukkerem: AddToScheme already covers PostUpgrade I think, same as in the other resources, but we could mirror App if that's preferred.

SatabdiG: No thats fine. I think App has a different contract.

apis/resources/v1alpha1/serviceinstance_types.go
SatabdiG: Oh then my bad!

SatabdiG (dismissed):
lgtm

lennardrother (approved):

PR #313, opened 2026-06-17
fix(serviceinstance): skip sharing reconciliation when sharedSpaces is nil

lennardrother:
Fix that managed service instances are unshared when `sharedSpaces` is not set when upgrading the provider version.


SatabdiG (approved):
:fire

PR #324, opened 2026-07-08
ci: fix runner for e2e workflow

n-losse:
used the wrong runner, wont be picked up


mpechkurov (approved):

n-losse:
admin override because this breaks the checks

PR #325, opened 2026-07-08
Fix skip logic

n-losse:
Fix erroneous skipping logic for e2e tests. 

Unfortunately, the job level skipping does not work if the job calls another workflow. So we need to pass an input to the e2e tests and skip the individual jobs in there, in order to satisfy the branch protection rule. 

Also, switch the collaborators logic back to the previous version, sind the collaborators API is not accessible without a PAT.

### DIFF
--- a/.github/workflows/pr-e2e.yaml
+++ b/.github/workflows/pr-e2e.yaml
@@ -44,7 +44,7 @@ jobs:
 
   approve:
     needs: [detect-non-doc-changes, get-author-association]
-    if: needs.detect-non-doc-changes.outputs.has-non-doc-changes == 'true'
+    if: needs.detect-non-doc-changes.outputs.has-only-doc-changes != 'true'
     runs-on: ubuntu-latest
     environment:
       name: ${{ needs.get-author-association.outputs.is-collaborator == 'true' && 'pr-e2e-no-approval' || 'pr-e2e-approval' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-e2e.yaml
```diff
@@ -52,9 +52,10 @@ jobs:
...
 
   run-e2e-test:
-    needs: approve
-    if: needs.approve.result == 'success'
+    needs: [approve, detect-non-doc-changes]
+    if: needs.approve.result == 'success' || needs.approve.result == 'skipped'
     uses: ./.github/workflows/e2e_test.yaml
     with:
       checkout-ref: ${{ github.event.pull_request.head.sha }}
+      skip: ${{ needs.detect-non-doc-changes.outputs.has-non-doc-changes != 'true' }}
```
sdischer-sap: Naming confuses me a little. Isn't it more like "contains-only-doc-changes" or something similar.

n-losse: true, i'll switch it around

sdischer-sap (dismissed):

mpechkurov (approved):
LGTM

n-losse:
Note for future self on the deprecation: seems it is going to keep working like this 

https://github.com/orgs/community/discussions/201388#discussioncomment-17602739

> Just a quick heads-up on that github changelog announcement:--> the deprecation is actually for the public Activity Events API (like when people query GET /events) not the actual webhook payloads that fill out github.event inside your action workflows. So it's definitely worth checking first whether github.event.pull_request.author_association is actually missing for you right now if it's still there, this fetch is just future-proofing rather than a required fix. But if you want to be completely safe against payload changes and still call the REST API like they recommended there's a way simpler workaround. The author_association field is still sitting right there on the Pull Request (or Issue) object if you fetch it straight from the REST API and since just reading a PR only requires basic access, the default GITHUB_TOKEN can handle it perfectly without any extra setup.

PR #327, opened 2026-07-15
feat(servicecredentialbinding): expose name in status

### LINKED ISSUE #323: [BUG] CF provider - ServiceCredentialBinding creates service keys with an unexpected random suffix
**Summary**

When creating an XSUAA ServiceCredentialBinding using the Crossplane Cloud Foundry Provider, the resulting service key in Cloud Foundry is created with an unexpected random suffix appended to the configured name.

For example, if the requested name is:

`auditlog-xsuaa-service-key-k8s
`
the actual service key created in Cloud Foundry is:

`auditlog-xsuaa-service-key-k8s-dvo6q
`
The expected behavior is that the service key name should exactly match the name specified in the ServiceCredentialBinding resource.

Expected Behavior

Given the following resource:

```
spec:
  forProvider:
    name: auditlog-xsuaa-service-key-k8s
```

the Cloud Foundry service key should be created as:

`auditlog-xsuaa-service-key-k8s
`
without any additional suffix.

Actual Behavior

The Cloud Foundry provider successfully creates the binding, but the service key appears in Cloud Foundry as:

`auditlog-xsuaa-service-key-k8s-dvo6q
`
where a random suffix is appended to the requested name.

The Crossplane resource reaches the Ready=True state, but the generated service key name does not match the configured name.

**Steps to Reproduce**

Create an XSUAA Service Instance.
Create the following ServiceCredentialBinding:
```
kind: ServiceCredentialBinding
spec:
  forProvider:
    name: auditlog-xsuaa-service-key-k8s
    type: key
    serviceInstanceRef:
      name: xsuaa-k8s-aws-cf-del101
```
Verify the created service key in the Cloud Foundry Cockpit.
Observed Result

**Service key name:**

auditlog-xsuaa-service-key-k8s-dvo6q
Expected Result

Service key name:

auditlog-xsuaa-service-key-k8s

**Additional Observations**

Creating the same service key manually through the Cloud Foundry Cockpit preserves the configured name and does not append a random suffix.
This suggests the suffix is introduced by the API flow used by the provider rather than being inherent XSUAA behavior.
The provider should use the supplied name as the service key name instead of allowing an auto-generated suffix.

**Environment**
Crossplane Cloud Foundry Provider: openmcp.common.repositories.cloud.sap/mirror/provider-cloudfoundry:v0.3.3 

**Question**

Is the provider intentionally using an API that generates unique names for service keys, or is this an implementation issue? Since both the Cloud Foundry Cockpit and cf create-service-key preserve the requested name, it would be expected that the provider behaves consistently by creating the service key with the exact configured name.

This would help us in refering service creds at places where secrets are not refered.

lennardrother:
This PR adds the name of the active and retired service credential bindings to the status alongside the already exposed GUIDs.

### DIFF
--- a/apis/resources/v1alpha1/servicecredentialbinding_types.go
+++ b/apis/resources/v1alpha1/servicecredentialbinding_types.go
@@ -109,6 +109,8 @@ type ServiceCredentialBindingStatus struct {
 type SCBResource struct {
 	// The GUID of the Cloud Foundry resource
 	GUID string `json:"guid,omitempty"`
+	// The name of the Cloud Foundry resource
+	Name string `json:"name,omitempty"`
 	// The date and time when the resource was created.
 	CreatedAt *metav1.Time `json:"createdAt,omitempty"`
 }
--- a/internal/clients/servicecredentialbinding/keyrotator.go
+++ b/internal/clients/servicecredentialbinding/keyrotator.go
@@ -56,7 +56,7 @@ func (r *SCBKeyRotator) RetireBinding(cr *v1alpha1.ServiceCredentialBinding, ser
 		}
 		cr.Status.AtProvider.RetiredKeys = append(cr.Status.AtProvider.RetiredKeys, &v1alpha1.SCBResource{
 			GUID:      serviceBinding.GUID,
-			Name:      ptr.Deref(serviceBinding.Name, ""),
+			Name:      ptr.Deref(serviceBinding.Name, ""), // only set during retirement; already retired keys are not backfilled
 			CreatedAt: &metav1.Time{Time: serviceBinding.CreatedAt},
 		})
 		return true
--- a/internal/clients/servicecredentialbinding/keyrotator.go
+++ b/internal/clients/servicecredentialbinding/keyrotator.go
@@ -9,6 +9,7 @@ import (
 	cfresource "github.com/cloudfoundry/go-cfclient/v3/resource"
 	"github.com/crossplane/crossplane-runtime/v2/pkg/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 
 	"github.com/SAP/crossplane-provider-cloudfoundry/apis/resources/v1alpha1"
 )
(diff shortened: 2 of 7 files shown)


n-losse (dismissed):

apis/resources/v1alpha1/servicecredentialbinding_types.go
```diff
@@ -109,6 +109,8 @@ type ServiceCredentialBindingStatus struct {
 type SCBResource struct {
 	// The GUID of the Cloud Foundry resource
 	GUID string `json:"guid,omitempty"`
+	// The name of the Cloud Foundry resource
+	Name string `json:"name,omitempty"`
```
mpechkurov, line 113: Should we extend unit tests to cover this field ?

internal/clients/servicecredentialbinding/keyrotator.go
```diff
@@ -55,6 +56,7 @@ func (r *SCBKeyRotator) RetireBinding(cr *v1alpha1.ServiceCredentialBinding, ser
 		}
 		cr.Status.AtProvider.RetiredKeys = append(cr.Status.AtProvider.RetiredKeys, &v1alpha1.SCBResource{
 			GUID:      serviceBinding.GUID,
+			Name:      ptr.Deref(serviceBinding.Name, ""),
```
mpechkurov: Should we add here the comment that this will only set on first retirement? 

I mean if the binding was already in "RetiredKeys" this will not get its Name.

apis/resources/v1alpha1/servicecredentialbinding_types.go
lennardrother: Thanks, done with https://github.com/SAP/crossplane-provider-cloudfoundry/pull/327/commits/51a42d64a23eb616d5361f278d5062aef2905811

internal/clients/servicecredentialbinding/keyrotator.go
lennardrother: Either this or I backfill it here: https://github.com/SAP/crossplane-provider-cloudfoundry/blob/51a42d64a23eb616d5361f278d5062aef2905811/internal/clients/servicecredentialbinding/keyrotator.go#L53

What do you think?

lennardrother: Nevermind. This would only add the name to the latest retired key. Let me add a comment instead.

lennardrother: Done via https://github.com/SAP/crossplane-provider-cloudfoundry/pull/327/commits/64e04dc137f1d013561061a8e5b1d7fb90741ee6

mpechkurov (approved):

Issue #332, opened 2026-08-03
[BUG]  Service key names get a random suffix even when rotation is not configured


SatabdiG:
## Problem
`newCreateOption` unconditionally runs the binding name through `randomName()`, appending a random 5-character suffix — even when `spec.rotation` is not set. 
Meanwhile `newListOptions`, used for the existence-check search, searches CF using the exact, unsuffixed `spec.Name`.

This mismatch means the existence-check can never find a binding it actually created, since the name in CF differs from the name being searched for.

This was previously identified as a contributing factor in the CIS credential-type orphaned-binding loop (search-by-name never matches → repeated `Create()` → repeated orphaned bindings), but the underlying naming issue was never independently actioned as its own fix.

## Reported by
Benjamin, via public Slack: noticed service keys getting random suffixes with
no `rotation:` block present.

## Expected behavior
The random suffix should only be appended when rotation is active (to avoid name collisions across rotated generations). For non-rotating keys, the name in CF should match `spec.Name` exactly, and the existence-check search should be able to find it.

## Suggested fix
Gate the `randomName()` call in `newCreateOption` on `spec.rotation` being set, matching the assumption already made in `newListOptions`.

PR #333, opened 2026-08-04
Random suffix rotation not configured

SatabdiG:
## Summary

No code change. This closes out the investigation into #332 (non-rotating service keys getting an unexpected random name suffix) by explaining why the current behavior is correct, and adds a doc-comment clarification so future users don't hit the same confusion.

## Investigation

PR #327 (merged) already addressed the actual user-facing problem: it exposes the real Cloud Foundry name via `status.atProvider.name` (and `status.atProvider.retiredKeys[].name` for rotated-out keys), so the
suffixed name is discoverable without needing `spec.Name` to match it.

## Change

`apis/resources/v1alpha1/servicecredentialbinding_types.go`: updated the doc comment on  `ServiceCredentialBindingParameters.Name`  

Relates to #332, relates to #327


xdj2e9:
I think this behavior was addressed in #323 and the solution was to expose the currently active key in the status field instead of removing the suffix (as this was not possible because of secret rotation; see https://github.com/SAP/crossplane-provider-cloudfoundry/issues/323#issuecomment-4934364553: `We apply the suffix regardless of whether rotation is enabled, so that enabling rotation later doesn't cause a name collision.`).

What do you think?

SatabdiG:
> I think this behavior was addressed in #323 and the solution was to expose the currently active key in the status field instead of removing the suffix (as this was not possible because of secret rotation; see [#323 (comment)](https://github.com/SAP/crossplane-provider-cloudfoundry/issues/323#issuecomment-4934364553): `We apply the suffix regardless of whether rotation is enabled, so that enabling rotation later doesn't cause a name collision.`).
> 
> What do you think?

good catch, I just saw Lennard's pr. But this doesn't solve the visibility problem for a user, who asks why we have random elements to the key. I would say we add a field comment atleast. 

What do you think?

xdj2e9:
> > I think this behavior was addressed in #323 and the solution was to expose the currently active key in the status field instead of removing the suffix (as this was not possible because of secret rotation; see [#323 (comment)](https://github.com/SAP/crossplane-provider-cloudfoundry/issues/323#issuecomment-4934364553): `We apply the suffix regardless of whether rotation is enabled, so that enabling rotation later doesn't cause a name collision.`).
> > What do you think?
> 
> good catch, I just saw Lennard's pr. But this doesn't solve the visibility problem for a user, who asks why we have random elements to the key. I would say we add a field comment atleast.
> 
> What do you think?

Documenting that behavior sounds good to me.

internal/clients/servicecredentialbinding/servicecredentialbinding.go
```diff
@@ -172,7 +172,10 @@ func newCreateOption(mg xpresource.Managed, forProvider v1alpha1.ServiceCredenti
 			return nil, errors.New(ErrNameMissing)
 		}
 
-		name := randomName(*forProvider.Name)
+		name := *forProvider.Name
+		if forProvider.Rotation != nil {
```
Kukkerem: Isn't this condition too narrow? Force-rotation can be triggered by the annotation alone, even without a `rotation` block, and the old key isn't removed from CF. For keys created under this change, the name is the plain spec name, so the replacement would reuse it, and CF would reject it as a duplicate. Shouldn't we check if a previous key still exists instead?

Maybe we could check RetiredKeys by adding this and checking it here with hasRetiredKeys(mg):

```go
func hasRetiredKeys(mg xpresource.Managed) bool {
   cr, ok := mg.(*v1alpha1.ServiceCredentialBinding)
   return ok && len(cr.Status.AtProvider.RetiredKeys) > 0
}
```

 RetiredKeys is persisted before Create runs, so it's set in time. WDYT?

Kukkerem (commented):
Looks great, one question only :fire:

internal/clients/servicecredentialbinding/servicecredentialbinding.go
SatabdiG: this wouldn't work as Patrick says. Also Lennard solved this issue a while ago. I forgot to revert the change yesterday. We can add better documentation, >.<

Kukkerem (approved):
LGTM 🔥

xdj2e9 (approved):

PR #334, opened 2026-08-05
fix(upgrade): support single-image provider layouts

### LINKED ISSUE #279: [BUG] Upgrade test framework does not support single-image provider build

gergely-szabo-sap:
## Summary

- Support local upgrade tests with single-image provider packages.
- Preserve compatibility with legacy two-image provider layouts.
- Skip pulling/loading optional controller images when absent.
- Add image-layout resolution and focused unit tests.
- Add local package-image loading support for kind.

## Validation

- Local upgrade test completed successfully.
- Added unit tests for single-image and two-image layouts.

### DIFF
--- a/test/upgrade/image_layout.go
+++ b/test/upgrade/image_layout.go
@@ -18,6 +18,10 @@ const (
 	firstSingleImageVersion = "1.0.1"
 )
 
+// firstSingleImageSemver is parsed once at init so invalid cutoffs fail
+// immediately rather than on first layout resolution.
+var firstSingleImageSemver = mustParseSemver(firstSingleImageVersion)
+
 type imageLayout struct {
 	packageImage    string
 	controllerImage *string
@@ -41,16 +45,19 @@ func resolveImageLayout(tag, localPackage string, localController *string, packa
 // tags (for example commit or channel tags) are assumed to refer to the
 // current single-image build; layout detection never depends on a failed pull.
 func isSingleImageVersion(tag string) bool {
-	threshold, err := semver.Parse(firstSingleImageVersion)
-	if err != nil {
-		panic(fmt.Errorf("invalid single-image version cutoff %q: %w", firstSingleImageVersion, err))
-	}
-
 	version, err := semver.Parse(strings.TrimPrefix(tag, "v"))
 	if err != nil {
 		return true
 	}
-	return version.GTE(threshold)
+	return version.GTE(firstSingleImageSemver)
+}
+
+func mustParseSemver(version string) semver.Version {
+	v, err := semver.Parse(version)
+	if err != nil {
+		panic(fmt.Errorf("invalid single-image version cutoff %q: %w", version, err))
+	}
+	return v
 }
 
 func providerInstallOptions(providerName string, layout imageLayout) xpenvfuncs.InstallCrossplaneProviderOptions {
--- a/test/upgrade/image_layout.go
+++ b/test/upgrade/image_layout.go
@@ -1,0 +1,95 @@
+package upgrade
+
+import (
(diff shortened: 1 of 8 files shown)


test/upgrade/upgrade_test_framework.go
```diff
@@ -274,11 +273,8 @@ func (b *CustomUpgradeTestBuilder) Feature() features.Feature {
 	return feature.Feature()
 }
 
-func loadPackages(fromTag, toTag string) (string, string) {
-	return test.LoadUpgradePackages(
-		fromTag, toTag,
-		fromPackage, toPackage,
-	)
+func loadPackages(fromTag, toTag string) (imageLayout, imageLayout) {
```
SatabdiG: Could we just call resolveImagePaths directly at the one call site and drop the wrapper?

test/upgrade/image_layout.go
```diff
@@ -0,0 +1,95 @@
...
+	controllerImage := fmt.Sprintf("%s:%s", controllerRepository, tag)
+	return imageLayout{packageImage: packageImage, controllerImage: &controllerImage}
+}
+
+// isSingleImageVersion uses the documented v1.0.1 layout cutoff. Non-semver
+// tags (for example commit or channel tags) are assumed to refer to the
+// current single-image build; layout detection never depends on a failed pull.
+func isSingleImageVersion(tag string) bool {
+	threshold, err := semver.Parse(firstSingleImageVersion)
+	if err != nil {
```
SatabdiG, line 49: nit: semver.Parse(firstSingleImageVersion) re-parses the constant on every call to isSingleImageVersion. Would hoisting to a package-level avoid the redundant work and the runtime panic path entirely (fails at init instead of first use)?

SatabdiG (commented):
👍 look good to me, some minor comments

test/upgrade/upgrade_test_framework.go
gergely-szabo-sap: Addressed in https://github.com/SAP/crossplane-provider-cloudfoundry/pull/334/commits/315e4dd0fb50b44532f05ead7acbc45a0abbf784

test/upgrade/image_layout.go
gergely-szabo-sap: Addressed in https://github.com/SAP/crossplane-provider-cloudfoundry/pull/334/commits/315e4dd0fb50b44532f05ead7acbc45a0abbf784

SatabdiG (approved):

